PR #97, opened 2021-01-07
Add auto-correct support

yishus:
This PR adds a simple auto-correct implementation.

This can be triggered with the new flag `-a`

An example of how a check can add corrections
```ruby
add_offense("Space missing after '{{'", node: node) do |corrector|
  corrector.insert_before(node, " ")
end
```

**TODO**
- [x] Print what offenses were fixed
- [x] Tests

Resolves https://github.com/Shopify/theme-check/issues/57

### DIFF
--- a/lib/theme_check/cli.rb
+++ b/lib/theme_check/cli.rb
@@ -10,6 +10,7 @@ module ThemeCheck
         -c, [--category]          # Only run this category of checks
         -x, [--exclude-category]  # Exclude this category of checks
         -l, [--list]              # List enabled checks
+        -a, [--auto-correct]      # Automatically fix offenses
         -h, [--help]              # Show this. Hi!
 
       Description:
@@ -25,6 +26,7 @@ module ThemeCheck
       command = :check
       only_categories = []
       exclude_categories = []
+      auto_correct = false
 
       args = argv.dup
       while (arg = args.shift)
@@ -37,6 +39,8 @@ module ThemeCheck
           exclude_categories << args.shift.to_sym
         when "--list", "-l"
           command = :list
+        when "--auto-correct", "-a"
+          auto_correct = true
         else
           path = arg
         end
@@ -45,6 +49,7 @@ module ThemeCheck
       @config = ThemeCheck::Config.from_path(path)
       @config.only_categories = only_categories
       @config.exclude_categories = exclude_categories
+      @config.auto_correct = auto_correct
 
       send(command)
     end
@@ -69,10 +74,10 @@ module ThemeCheck
       if theme.all.empty?
         raise Abort, "No templates found.\n#{USAGE}"
       end
-      analyzer = ThemeCheck::Analyzer.new(theme, @config.enabled_checks)
+      analyzer = ThemeCheck::Analyzer.new(theme, @config.enabled_checks, @config.auto_correct)
       analyzer.analyze_theme
-      ThemeCheck::Printer.new.print(theme, analyzer.offenses)
-      raise Abort, "" if analyzer.offenses.any?
+      ThemeCheck::Printer.new.print(theme, analyzer.offenses, @config.auto_correct)
+      raise Abort, "" if analyzer.uncorrectable_offenses.any?
     end
   end
 end
--- a/lib/theme_check/template.rb
+++ b/lib/theme_check/template.rb
@@ -42,6 +42,11 @@ module ThemeCheck
       lines[line - 1].strip
     end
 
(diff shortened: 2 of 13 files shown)


lib/theme_check/template.rb
```diff
@@ -58,6 +58,15 @@ def root
       parse.root
     end
 
+    def update!(line, value)
```
macournoyer, line 61: I would rename to `replace_line` or `replace(line:, ...)`

lib/theme_check/template.rb
```diff
@@ -58,6 +58,15 @@ def root
...
+      lines[line - 1] = value
+      @source = lines.join("\n")
+    end
+
+    def write
+      File.write(path, source)
```
macournoyer, line 67: I think we should track when files are changed (maybe a ivar?) and not re-write the file when it hasn't change.

This way, it won't touch the files needlessly.

macournoyer (commented):
Looking real good! Nice work! And very clever approach to replacing the line 👏 

I would also be nice if we had a test helper to help test correctors easily.

lib/theme_check/template.rb
yishus: ah that's a good point

lib/theme_check/template.rb
```diff
@@ -58,13 +58,20 @@ def root
       parse.root
     end
 
-    def update!(line, value)
-      lines[line - 1] = value
-      @source = lines.join("\n")
+    def corrected_lines
```
macournoyer, line 61: Can we keep the "corrected" word out of `Template`? I don't think it needs to have any knowledge of autocorrection.

And technically they are not really corrected, unless `correct_line` has been called before.

I would rename to `lines`, and rename `correct_line` to `update_line`.

yishus: Ah for this I didn't want to amend the existing `lines` ivar in template, instead creating a duplicate that holds the amended lines. Although for now I don't see a problem amending lines directly, it feels a little iffy to dirty the source of truth. 

Lemme work through this naming, although would love to get your thoughts on amending the lines ivar directly

lib/theme_check/template.rb
```diff
@@ -58,6 +66,17 @@ def root
       parse.root
     end
 
+    def replace_line(line)
```
macournoyer, line 69: This is confusing. According to the method name, I would expect this method to actually replace the line.

Either I would expose an `updated!` method, or I would rename this method to make it clearer it is not doing the replacement.

W/ the `updated!` method:
```
line = template.full_line(line_number)
line.insert(...)
template.updated!
```

test/corrector_test.rb
```diff
@@ -0,0 +1,62 @@
...
+    )
+  end
+
+  def test_insert_after_adds_suffix
+    node = stub(
+      template: @theme["templates/index"],
+      line_number: 2,
+      range: [4, 8]
+    )
+    config = ThemeCheck::Corrector.new(template: node.template)
```
macournoyer, line 21:
```suggestion
    corrector = ThemeCheck::Corrector.new(template: node.template)
```

macournoyer (approved):
Just some internal method names I think are a bit confusing. But the API is really good 👏

lib/theme_check/template.rb
yishus: that's a really good idea. I wanted to make use of the destructive nature of insert to update full line without having to duplicate the string. But was stuck at where the call `@updated = true` should reside. Having a `update!` method would do the trick

test/corrector_test.rb
yishus: haha a bad case of copy and paste, 🙇

lib/theme_check/cli.rb
```diff
@@ -10,6 +10,7 @@ class Abort < StandardError; end
         -c, [--category]          # Only run this category of checks
         -x, [--exclude-category]  # Exclude this category of checks
         -l, [--list]              # List enabled checks
+        -a, [--auto-correct]      # Automatically fix offenses
```
charlespwd, line 13: I believe the standard flag for this is `-f, [--fix]`. 

That's how it's called in eslint. And that's also how this is called in w0rp/ale (for vim)

charlespwd: But then again rubocop calls this auto-correct 🙃

yishus: Ah yea I was deciding between the 2 but went ahead with rubocop’s naming, because Ruby \_(ツ)_/¯

PR #101, opened 2021-01-12
Add corrector to double braces space check

yishus:
Adds corrector to the double braces space checks so that they can be auto-corrected. Added also a fix to `corrector.rb` to update a node's markup on replace.

I'm hesitant to add the corrector to space check for liquid tags because we are able to check for missing or extra spaces before the closing brace but not after the opening brace.


macournoyer (commented):
Looks really cool! Do you have some ideas on how we could test this?

Re-read the files from disk, and assert their content is what we expect maybe?

macournoyer (approved):

Issue #8, opened 2020-11-26
Check for deprecated filters


macournoyer:
See https://shopify.dev/docs/themes/liquid/reference/filters/deprecated-filters

PR #105, opened 2021-01-18
Check for deprecated filters

tskorupa:
Implements https://github.com/Shopify/theme-check/issues/8

I chose to write a new filter, instead of extending the UndefinedFilter, because deprecated filters continue to work to support existing themes. Those filters are no longer supported. Additionally, I wanted to make this check return a severity of type suggestion.


macournoyer (approved):
With suggested alternatives!!! 🔥 👏

PR #112, opened 2021-01-20
[Language Server] Send empty diagnostics to flush errors

### LINKED ISSUE #111: Language server - errors remain after fix
After fixing all errors on a template, VSCode will continue to highlight them as problems.

Before
Screenshot 2021-01-20 at 2 58 23 PM

After
Screenshot 2021-01-20 at 2 58 36 PM

yishus:
Resolves #111 

Error occurs when a template moves from having offences to none. Currently we are not publishing diagnostics if a template contains no offence, however the extension would also not get the updated state.

### DIFF
--- a/lib/theme_check/language_server/handler.rb
+++ b/lib/theme_check/language_server/handler.rb
@@ -47,21 +47,33 @@ module ThemeCheck
         log("Checking #{config.root}")
         analyzer.analyze_theme
         log("Found #{theme.all.size} templates, and #{analyzer.offenses.size} offenses")
-        send_offenses(analyzer.offenses)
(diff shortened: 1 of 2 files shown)


charlespwd (approved):
Must have been lost in the refactor. Good catch that's exactly it.

macournoyer:
I think there's still an issue if you fix all the errors in a file, but there are still in another file.

I think you need to send empty diagnostics for all files that were previously reported.

yishus:
@macournoyer that makes sense. okay updated that we send empty diagnostics for all other template files

macournoyer (approved):
Awesome!

PR #113, opened 2021-01-20
Remove direct link to code editor support

macournoyer:
We'll put it back when it is available outside.


charlespwd (approved):

macournoyer:
Also added a "coming soon" note to the Wiki

Issue #118, opened 2021-01-21
Unable to execute without an exception


Tiberriver256:
## Recreation Steps

```sh
gem install theme-check
cd /my/project
theme-check
```

I see the following exception
```
Traceback (most recent call last):
        8: from /usr/local/bin/theme-check:23:in `<main>'
        7: from /usr/local/bin/theme-check:23:in `load'
        6: from /var/lib/gems/2.5.0/gems/theme-check-0.2.0/exe/theme-check:4:in `<top (required)>'
        5: from /usr/lib/ruby/2.5.0/rubygems/core_ext/kernel_require.rb:59:in `require'
        4: from /usr/lib/ruby/2.5.0/rubygems/core_ext/kernel_require.rb:59:in `require'
        3: from /var/lib/gems/2.5.0/gems/theme-check-0.2.0/lib/theme_check.rb:8:in `<top (required)>'
        2: from /var/lib/gems/2.5.0/gems/theme-check-0.2.0/lib/theme_check.rb:8:in `require_relative'
        1: from /var/lib/gems/2.5.0/gems/theme-check-0.2.0/lib/theme_check/liquid_check.rb:2:in `<top (required)>'
/var/lib/gems/2.5.0/gems/theme-check-0.2.0/lib/theme_check/liquid_check.rb:2:in `require_relative': /var/lib/gems/2.5.0/gems/theme-check-0.2.0/lib/theme_check/parsing_helpers.rb:9: syntax error, unexpected .., expecting ']' (SyntaxError)
...       yield scanner.matched[..-2]
...                             ^~
```

## Ruby Versions Tested
```
ruby 2.5.1p57 (2018-03-29 revision 63029) [x86_64-linux-gnu]
ruby 2.6.6p146 (2020-03-31 revision 67876) [x64-mingw32]
```

macournoyer:
We do require [Ruby 2.7+ at the moment](https://github.com/Shopify/theme-check#requirements). I'll see if running on 2.6 is an easy fix.

Tiberriver256:
Ah gotcha. No worries I just updated Ruby and it worked fine. I'm not too familiar with Ruby and the linux one was just the default that came from running `sudo apt install ruby-full` and the Windows one I had installed awhile back.

Thank you!

PR #120, opened 2021-01-21
Add Ruby 2.7 as a gem requirement

macournoyer:
To avoid confusing errors such as #118


charlespwd (approved):
LGTM. Should probably add that requirement in the brew one as well. Would need to check how.

PR #122, opened 2021-01-22
Ignore node on assign type 

akinshonibare:
Came across an issue with Space checking between `,:` characters while adding `|`
On an assign tag the check returns the error twice as seen below
since the assign node markup returns the whole content including the variable

assign node markup : `assign max_width = height | times:  image.aspect_ratio`
variable node markup: `height | times:  image.aspect_ratio`

error thrown
```
+"Too many spaces after ':' at templates/index.liquid:1
+Too many spaces after ':' at templates/index.liquid:1"
```

This pr ignores the check if node type is `assign`;

### DIFF
--- a/lib/theme_check/checks/space_inside_braces.rb
+++ b/lib/theme_check/checks/space_inside_braces.rb
@@ -11,6 +11,7 @@ module ThemeCheck
 
     def on_node(node)
       return unless node.markup
+      return if :assign == node.type_name
 
       outside_of_strings(node.markup) do |chunk|
         chunk.scan(/([,:])  +/) do |_match|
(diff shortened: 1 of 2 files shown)


macournoyer (changes requested):
Could you add a test case to repro?

I assume something w/:

```
{% assign max_width = height | times: image.aspect_ratio %}
```

would reproduce it, correct?

akinshonibare:
> Could you add a test case to repro?

Screen Shot 2021-01-25 at 9 18 33 AM

Same test without skipping `assign` nodes

macournoyer (approved):

Issue #79, opened 2020-12-13
Add ability to disable checks with comments


macournoyer:
We need a way to disable individual checks, or all checks for a block of markup.

Something like https://eslint.org/docs/2.13.1/user-guide/configuring#disabling-rules-with-inline-comments or Rubocop's:
```ruby
# rubocop:disable RuleByName
This is a long line 
# rubocop:enable RuleByName
```

### Disabling all checks

```liquid
{% comment %}theme-check-disable{% endcomment %}
...
{% comment %}theme-check-enable{% endcomment %}
```

### Disabling individual checks

```liquid
{% comment %}theme-check-disable UnusedAssign{% endcomment %}
...
{% comment %}theme-check-enable UnusedAssign{% endcomment %}
```

### How to implement
We do have access to the comments in the `LiquidCheck`s.

```ruby
def on_comment(node)
  node.value.nodelist.join # <= Should have the content of the comment tag
end
```

### Adding a check for missing `theme-check-enable` comments

Introducing this also introduces the risk of someone forgetting to put a `theme-check-enable` comment to close the ignored block. Lets create a check for this! Each `theme-check-disable` comment, must have a matching `theme-check-enable`.

PR #131, opened 2021-01-25
Allow checks to be disabled with comments

jplhomer:
This PR adds support for disabling checks using a comment syntax similar to that of Rubocop:

```liquid
{% comment %}theme-check-disable{% endcomment %}
{% assign x = 1 %}
{% comment %}theme-check-enable{% endcomment %}
```

Disable a specific check by including it in the comment:

```liquid
{% comment %}theme-check-disable UnusedAssignCheck{% endcomment %}
{% assign x = 1 %}
{% comment %}theme-check-enable UnusedAssignCheck{% endcomment %}
```

Disable multiple checks by including them as a comma-separated list:

```liquid
{% comment %}theme-check-disable UnusedAssignCheck,SpaceInsideBracesCheck{% endcomment %}
{%assign x = 1%}
{% comment %}theme-check-enable UnusedAssignCheck,SpaceInsideBracesCheck{% endcomment %}
```

Also adds a new rule: `MissingEnableComment` to show the user when they've forgotten to re-enable a check.

Fixes https://github.com/Shopify/theme-check/issues/79

## How it works

- Adds a `DisabledChecks` helper
- Inside `Visitor`, the helper tracks enable/disable comments when the current node is a comment
- Keeps track of a list of disabled checks (and whether all checks are disabled)
- Adds an `except_disabled` method to `Checks` collection to filter out currently disabled checks

In order to support the new `MissingEnableComment` rule, we need a way to add an exception to disabled checks so the rule itself is not ignored.

This adds a new syntax to the `Check` interface:

```rb
always_enabled!
```

Which checks can use to become resistant to comment disabling. I anticipate to be a private API and not likely used by any other checks, but it's possible that a consumer may wish to prevent users from disabling certain checks for a "strict mode."

## Other things considered

Rubocop uses a more detailed [`CommentConfig`](https://github.com/rubocop-hq/rubocop/blob/7bf9d469c50763573f8380c6d8ac0907589fdd3c/lib/rubocop/comment_config.rb) utility, which tracks line number ranges of disable/enable comments. It also supports inline disabling with a single comment.

I didn't take this approach because I'm not certain whether the Liquid AST supports line numbers in the way we anticipate. However, I'm definitely open to alternative implementations!


lib/theme_check/checks/missing_enable_comment.rb
```diff
@@ -0,0 +1,30 @@
+# frozen_string_literal: true
+module ThemeCheck
+  class MissingEnableComment < LiquidCheck
+    severity :error
+
+    # Don't allow this check to be disabled with a comment,
+    # as we need to be able to check for disabled checks.
+    always_enable!
```
macournoyer, line 8: I think changing to `disableable false` (or `can_disable`) would make it clearer what this does.

lib/theme_check/checks.rb
```diff
@@ -8,5 +8,18 @@ def call(method, *args)
         end
       end
     end
+
+    # Return a collection of checks except for disabled checks represented
+    # as a DisabledChecks instance — or `nil`, if all checks are meant to be disabled.
+    # Still returns checks marked as `always_enabled!`
+    def except_disabled(disabled_checks = nil)
```
macournoyer, line 15: It's not clear to me what the method arg does. Maybe it would be simpler if it was split in two methods: one the same as `disabled_check=nil`, and the other when present.

test/checks/missing_enable_comment_test.rb
```diff
@@ -0,0 +1,62 @@
...
+class MissingEnableCommentTest < Minitest::Test
+  def test_always_enabled_by_default
+    assert(ThemeCheck::MissingEnableComment.new.always_enabled?)
+  end
+
+  def test_no_default_noops
+    offenses = analyze_theme(
+      ThemeCheck::MissingEnableComment.new,
+      "templates/index.liquid" => <<~END,
+        {% theme-check-disable %}
```
macournoyer, line 14: This is a Liquid syntax error (because the tag doesn't exist). The `SyntaxError` check would report it. Is this what you wanted to test?

test/checks/missing_enable_comment_test.rb
```diff
@@ -0,0 +1,62 @@
+# frozen_string_literal: true
+require "test_helper"
+require "minitest/focus"
+
+class MissingEnableCommentTest < Minitest::Test
```
macournoyer, line 5: Would it make sense to also look for `theme-check-enable` with missing `theme-check-disable`?

test/disabled_checks_test.rb
```diff
@@ -0,0 +1,150 @@
...
+  def test_ignore_all_checks
+    template = parse_liquid(<<~END)
+      {% comment %}theme-check-disable{% endcomment %}
+      {% assign x = 'hello' %}
+      {% comment %}theme-check-enable{% endcomment %}
+      hello
+    END
+    @visitor.visit_template(template)
+
+    assert_equal([
```
macournoyer, line 22: I think asserting assign it not visited instead would make it clearer and less fragile to future changes:

```
refute_includes(@tracer_one.calls, :on_assign)
```

Same for other ones bellow.

README.md
```diff
@@ -87,3 +88,29 @@ TemplateLength:
...
+```liquid
+{% comment %}theme-check-disable{% endcomment %}
+{% assign x = 1 %}
+{% comment %}theme-check-enable{% endcomment %}
+```
+
+Disable a specific check by including it in the comment:
+
+```liquid
+{% comment %}theme-check-disable UnusedAssignCheck{% endcomment %}
```
macournoyer, line 105:
```suggestion
{% comment %}theme-check-disable UnusedAssign{% endcomment %}
```

test/disabled_checks_test.rb
```diff
@@ -0,0 +1,150 @@
+# frozen_string_literal: true
+require "test_helper"
+
+class DisabledChecksTest < Minitest::Test
+  def setup
+    @tracer_one = TracerCheck.new
+    @tracer_two = TracerCheck.new
+    @tracer_one.stubs(:code_name).returns('TracerOneCheck')
```
macournoyer, line 8: Instead of stubbing you could create two real classes nested under `DisabledChecksTest`:

```
class TrackerOne < TracerCheck; end
class TrackerTwo < TracerCheck; end
```

I feel it would avoid this test knowing about the internals for `Checks`.

macournoyer (commented):
Looking good! 👍

charlespwd:
Can we also make it so it ignores the whitespace? Kind of wished there was spacing there:

```html
{% comment %} theme-check-disable {% endcomment %}

<!-- Could probably put that at the top of a document -->
{% comment %}
theme-check-disable Foo,Bar,Baz
{% endcomment %}
```

Also do we _need_ a `theme-check-enable`? It's a common thing to do disable rules for entire files with eslint (with only a set of disabled rules at the top of the file)

macournoyer:
> Also do we _need_ a `theme-check-enable`? It's a common thing to do disable rules for entire files with eslint (with only a set of disabled rules at the top of the file)

Ah good point! But then it needs to be at the top. So the check should be, `theme-check-enable` is not required if `theme-check-disable` is in the first line of the file?

```
# Bad
blablablabla

{% comment %}theme-check-disable{% endcomment %}

# Good
{% comment %}theme-check-disable{% endcomment %}

Blablabla
```

test/checks/missing_enable_comment_test.rb
jplhomer: Whoops! Great catch.

lib/theme_check/checks.rb
jplhomer: I tried clearing this up by splitting out the methods as suggested. Still feels a bit off to me, so please let me know if you have suggestions!

test/checks/missing_enable_comment_test.rb
jplhomer: I like the idea. I tried implementing this, but I'm struggling to come up with a good mechanism for checking this.

My naive approach was to keep a simple counter (+/-) based on # of disable/enable statements. If the count is > 0 at the end of the document, it means there's a superfluous `enable` comment.

However, that doesn't work when you have multiple sets of checks, or you granularly enable certain checks.

```liquid
{% comment %}theme-check-disable TracerCheck, AnotherCheck{% endcomment %}
Stuff
{% comment %}theme-check-enable TracerCheck{% endcomment %}
More stuff
{% comment %}theme-check-enable AnotherCheck{% endcomment %}
```

Finally, it seems like we ought to allow users to blanket-enable all checks at any given point, meaning we can't even rely upon the # of checks re-enabled granularly:

```liquid
{% comment %}theme-check-disable TracerCheck, AnotherCheck, MoreChecks{% endcomment %}
Stuff
{% comment %}theme-check-enable TracerCheck{% endcomment %}
Blanket re-enable:
{% comment %}theme-check-enable{% endcomment %}
```

I'm OK passing on this for now and considering an implementation down the line if we find one. Thoughts?

macournoyer: Yes, you're right. Should be in another PR. Can you open an issue afterwards so we don't forget about this?

macournoyer (approved):
Killer work! 👏

PR #134, opened 2021-01-27
Add an option to ignore objects with a given prefix in UndefinedObject

### LINKED ISSUE #133: [Bug] Undefined object error when rendering snippet with optional parameters
## Problem

Sometimes in themes, we will render a snippet using "optional" parameters.

### Example:

``` liquid
 {% render 'product-card', media_size: section.settings.product_image_ratio %}
```

**The corresponding snippet can accept up to 5 parameters:**

``` liquid
{% comment %}
    Renders a product card

    Accepts:
    - product_card_product: {Object} Product Liquid object (optional)
    - show_vendor: {Boolean} Show the product's vendor depending on the section setting (optional)
    - media_size: {String} Size of the product image card. Values are "square" and "portrait". Default is "square" (optional)
    - show_full_details: {Boolean} Show the link to full details (optional). Default to hide
    - center_align_text: {Boolean} Center all the text contained inside. Default is to left-align (optional)

    Usage:
    {% render 'product-card', show_vendor: section.settings.show_vendor %}
{% endcomment %}
```

These parameters are all meant to be optional.  For example, when a section is being rendered in its onboarding state there may not be any product passed to the product card snippet.

**`theme-check` throws the following errors because the objects are unidentified in that circumstance:**

```
Missing argument `product_card_product`theme-check(UndefinedObject)
Missing argument `show_full_details`theme-check(UndefinedObject)
Missing argument `center_align_text`theme-check(UndefinedObject)
Missing argument `show_vendor`theme-check(UndefinedObject)
```

## Expected

For "optional" parameters to be recognized as such and not throw an error in cases like this.

One idea could be just a code standard where we prefix optional parameters somehow -  `o_product_card_product` - to have it ignore the `UndefinedObject` check.

macournoyer:
Fixes #133

### DIFF
--- a/config/default.yml
+++ b/config/default.yml
@@ -58,6 +58,8 @@ MissingRequiredTemplateFiles:
 
 UndefinedObject:
   enabled: true
+  # Objects with this prefix will be ignored
+  optional_prefix: o_
 
 RequiredDirectories:
   enabled: true
--- a/lib/theme_check/checks/undefined_object.rb
+++ b/lib/theme_check/checks/undefined_object.rb
@@ -49,8 +49,9 @@ module ThemeCheck
       end
     end
 
-    def initialize
+    def initialize(optional_prefix: nil)
       @files = {}
+      @optional_prefix = optional_prefix
     end
 
     def on_document(node)
@@ -85,8 +86,11 @@ module ThemeCheck
     end
 
     def on_variable_lookup(node)
+      variable_name = node.value.name
+      return if @optional_prefix && variable_name.starts_with?(@optional_prefix)
+
       @files[node.template.name].add_variable_lookup(
-        name: node.value.name,
+        name: variable_name,
         node: node,
       )
     end
--- a/test/checks/undefined_object_test.rb
+++ b/test/checks/undefined_object_test.rb
@@ -316,4 +316,17 @@ class UndefinedObjectTest < Minitest::Test
       Undefined object `email` at templates/index.liquid:1
     END
   end
+
+  def test_ignores_optional
+    offenses = analyze_theme(
+      ThemeCheck::UndefinedObject.new(optional_prefix: "o_"),
+      "templates/index.liquid" => <<~END,
+        {% render 'product' %}
+      END
+      "snippets/product.liquid" => <<~END,
+        {{ o_undefined }}
+      END
+    )
+    assert_offenses("", offenses)
+  end
 end


config/default.yml
```diff
@@ -58,6 +58,8 @@ MissingRequiredTemplateFiles:
 
 UndefinedObject:
   enabled: true
+  # Objects with this prefix will be ignored
+  optional_prefix: o_
```
charlespwd, line 62: * Is this an existing convention or a new feature?
* Should this be convered by something like `disable-theme-check UndefinedObject:foo-bar`?

macournoyer: It was suggested by @tyleralsbury in #133. It's the only way I could think of to support detecting undefined objects, or else we have to drop the whole check.

That sounds like a got idea for the comment. Do you know if eslint allows something similar?

tyleralsbury: As for whether it's a convention or something new - in this case, it's something new, though it's more or less based on established patterns like dangling underscores for private methods in other languages.

---

The comment idea is an interesting approach as well.  In ESLint you have two options - you can either disable it for a line or for the entire file, which might be good for cases like this, so maybe:

`disable-next-line-theme-check UndefinedObject:foo_bar`

and

`disable-file-theme-check UndefinedObject:foo_bar`

To allow for both options.

Being able to do this for all of the things `theme-check` looks for would be a nice way of allowing the developer to control which rules they want to care about for a particular file or line to account for all cases that we haven't thought of - if the developer is getting a report from `theme-check` that they think is wrong, they can just disable it.

Ideally, this would also come with a "quick fix" solution in the VS Code plugin.  When you have an issue like that in ESLint, you can click "quick fix" and one of the options is "add disable-eslint ... ` for that particular line.

macournoyer: @tyleralsbury good news 😄  => https://github.com/Shopify/theme-check#disable-checks-with-comments (not yet released on brew)

So the snippet template using optional variables would have this at the top:
```
{% comment %}
theme-check-disable UndefinedObject:optional_var1
theme-check-disable UndefinedObject:optional_var2
{% endcomment %}
``` 

Sounds good?

tyleralsbury: Yeah I think this is a really good solution - it accounts for this case and any other possible case that we didn't think of specifically.  Very scalable. 👍

tskorupa: Most of the optional variables in templates could be assigned at the top unless they are passed in. I'm not sure if this is possible in liquid;
```
assign optional_variable = optional_variable || nil
```

This would avoid introducing a linter language to the templates.

macournoyer: Oh yeah that's really good @tskorupa !!

But can't use expression in assign, only filters I think. But since variables default to `nil` it could be:
```
{% assign optional = optional %}
```
It does look weird :/

macournoyer: This would work too, more explicit:
```
{% assign optional = optional | default: nil %}
```

macournoyer:
I'm closing this in favor using `{% assign optional = optional | default: nil %}` to define optional vars.

I'll document in the ticket, but we might also need a way to educate users about this. Not sure the error message is the right place...

tskorupa:
> I'll document in the ticket, but we might also need a way to educate users about this. Not sure the error message is the right place...

Let me know if the error belongs to the snippet instead, I can modify it to say the likes of "undefined variable FOO, if it's an optional variable, you can add `{% assign optional = optional | default: nil %}` at the top of the snippet".

Even better, we can make that fix automatically if the developer wants it so.

PR #138, opened 2021-01-29
Add config to ignore files / directories

### LINKED ISSUE #128: Add config to ignore directories
In `.theme-check.yml`, we need a why to ignore directories and default it to

```
- node_modules
```

Since we scan .json files, this dir can result in thousands of files being scanned.

macournoyer:
In `.theme-check.yml`
```
ignore:
  - node_modules/*
  - docs/*.json
```

Will ignore all files matching any of those patterns (matched using [`fnmatch`](https://ruby-doc.org/core-2.5.0/File.html#method-c-fnmatch)).

Defaults to `["node_modules/*"]`.

Fixes #128 

I also added a bit more validation to `Config`:
- Check for type, by comparing to default
- Check for unknown configs, by comparing to default
- Ignore custom checks, but validate its config is a Hash

### DIFF
--- a/config/default.yml
+++ b/config/default.yml
@@ -3,7 +3,7 @@ root: .
 require: []
 
 ignore:
-  - node_modules
+  - node_modules/*
 
 ConvertIncludeToRender:
   enabled: true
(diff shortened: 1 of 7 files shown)


yishus (approved):
thanks for all the improvements on config in addition to adding ignored directories! this looks good 🙌

PR #145, opened 2021-02-08
Add InMemory{Theme,JsonFile,Template}

charlespwd:
Add an alternative to file system themes so that theme-check can be run on a hash of files -> content.

This is the first step in making Theme Check work in non-file-system contexts. A server perhaps?

Intentionally not allowing requires for configs because that could turn into a security issue.

Added the tests in the base test files to make sure that new features to the file-system interfaces would also be copied over to the in-memory ones.


test/theme_test.rb
```diff
@@ -3,49 +3,75 @@
...
   end
 
   def test_default_locale
-    assert_equal("en", @theme.default_locale)
+    @themes.each do |theme|
+      assert_equal("en", theme.default_locale)
+    end
   end
 
   def test_ignore
```
charlespwd, line 77: I didn't port this test since I don't think it makes sense. This is strictly testing the constructor of Theme (not InMemoryTheme) and handles things that would be managed by the input files hash to InMemoryTheme.new.

jplhomer (approved):
🙌

macournoyer (approved):
Conceptually I would have expected the inheritance to be the other way around, ie.: A `Template` loaded from disk is a specialization of an `InMemoryTemplate`, because it does less.

But, I think it would be better to extract everything related to storage in its own class (`InMemoryStore`, `FileSystemStore`?) instead, so we don't have to extend or add/remove methods from multiple classes.

Not saying it should be done in this PR.

charlespwd:
OK I'll merge this as is and will make the refactor as part of a different PR (working on it).

PR #146, opened 2021-02-08
Add ParserBlockingJavaScript Check

### LINKED ISSUE #78: Opinionated HTML rules
As I understand it, the AST doesn't really recognize the HTML as HTML. Am I wrong here? Looks like there would be some gain to be had by linting HTML as well. But might be complicated. 

What I'd really like is some opinionated checks that prevent stuff like: 

* <script> without `async` or `defer`
*  without srcset
* ...?

But really it's mostly about <script> without async or defer. That should not exist. And theme-check seems like a logical place to suggest/enforce this. 

Also perhaps an opportunity to add magic comments to disable rules? À la `// eslint-disable-line camel-case`

charlespwd:
Say goodbye to slow sites. Add friction to under performing HTML.

Fixes #78


lib/theme_check/checks/parser_blocking_javascript.rb
```diff
@@ -0,0 +1,50 @@
...
+        message: "The script_tag filter is parser-blocking, use a script tag with the async or defer attribute for better performance",
+        regex: SCRIPT_TAG_FILTER,
+      )
+    end
+
+    # The trickiness here is matching on scripts that are defined on
+    # multiple lines (or repeat matches). This makes the line_number
+    # calculation a bit weird. So instead, we traverse the string in
+    # a very imperative way.
+    def record_offences_from_regex(regex: nil, message: nil)
```
jplhomer, line 34: For consistency with `add_offense`

```suggestion
    def record_offenses_from_regex(regex: nil, message: nil)
```

lib/theme_check/checks/parser_blocking_javascript.rb
```diff
@@ -0,0 +1,50 @@
...
+
+    private
+
+    def record_offences
+      record_offences_from_regex(
+        message: "Missing async or defer attribute on script tag",
+        regex: PARSER_BLOCKING_SCRIPT_TAG,
+      )
+      record_offences_from_regex(
+        message: "The script_tag filter is parser-blocking, use a script tag with the async or defer attribute for better performance",
```
jplhomer, line 25:
```suggestion
        message: "The script_tag filter is parser-blocking. Use a script tag with the async or defer attribute for better performance",
```

test/checks/parser_blocking_javascript_test.rb
```diff
@@ -0,0 +1,113 @@
...
+
+  def test_script_tag_filter
+    offenses = analyze_theme(
+      ThemeCheck::ParserBlockingJavaScript.new,
+      "templates/index.liquid" => <<~END,
+        {{ 'foo.js' | asset_url | script_tag }}
+      END
+    )
+    assert_offenses(<<~END, offenses)
+      The script_tag filter is parser-blocking, use a script tag with the async or defer attribute for better performance at templates/index.liquid:1
```
jplhomer, line 69:
```suggestion
      The script_tag filter is parser-blocking. Use a script tag with the async or defer attribute for better performance at templates/index.liquid:1
```

jplhomer (approved):
😍 Dang, this is really awesome.

macournoyer (commented):
I don't think it's a good idea to start using regexes. How about `<script>`s inside comments? Inside strings? How about `< string ...`? We should use a real HTML parser. Like we do for Liquid and JSON.

We already have an HTML parser in the project: https://github.com/Shopify/theme-check/blob/master/lib/theme_check/checks/valid_html_translation.rb#L26-L25.

We could introduce a new `HtmlCheck` that has `on_...` callback methods like `LiquidCheck` & `JsonCheck`.

charlespwd:
> I don't think it's a good idea to start using regexes. How about `<script>`s inside comments? Inside strings? How about `< string ...`? We should use a real HTML parser. Like we do for Liquid and JSON.
> 
> We already have an HTML parser in the project: https://github.com/Shopify/theme-check/blob/master/lib/theme_check/checks/valid_html_translation.rb#L26-L25.
> 
> We could introduce a new `HtmlCheck` that has `on_...` callback methods like `LiquidCheck` & `JsonCheck`.

I'm not sure that's possible. How can we make an AST from something like this?

```liquid
<html>
<head>
  {% if some_property %}<script src=""></script>{% endif %}
  <scr{% if some_property %}ript{% end %}></script>
  <script{% if async %} async{% end %}{% if defer %} defer{% end %}></script>
</head>
</html>
```

This is technically valid liquid but I don't think you could parse an AST out of this since you'd need to context switch between HTML and liquid. Plus what if the liquid makes the tag insertion conditional? For our linter's purposes, is it there or is it not? What about conditional attributes on a tag?

I know the regex based approach is not ideal and is loopy but the intention is not of blocking loop holes. It's more about preventing you from doing something you were not even aware of / or preventing you from making a mistake.

If your goal is to bypass the rule, you'd have an easier time using {% comment %}theme-check-disable{% endcomment %}

I would even go as far as saying that if the script is parser blocking in a comment and we tell you, then you should just remove it or add a comment that makes it OK.

macournoyer:
It is static analysis so technically is should analyze the code in all the conditional branches:

```liquid
{% if some_property %}
<!-- Would analyze this, as an independent HTML fragment -->
<script src=""></script>
{% else %}
<!-- ... and this, as another independent HTML fragment -->
<script src=""></script>
{% endif %}
```

However for those it would indeed be very complex:

```liquid
  <scr{% if some_property %}ript{% end %}></script>
  <script{% if async %} async{% end %}{% if defer %} defer{% end %}></script>
```

But that would also be tricky w/ the regexes, no?

Do you have other HTML checks in mind?

charlespwd:
> But that would also be tricky w/ the regexes, no?

Like I said, the idea is not really to prevent loopholes. It's to make sure that the basic stuff doesn't go through. Which the regex does kind of well.

Other checks that I have in mind that I may, definitely, implement:
- File size limits on JS bundles
- File size limits on CSS bundles

Stuff I'm thinking of maybe implementing:
- Using srcset on `` tags
- Google optimize warning :p

Kicking the can:
- Loading assets from domains other than shopify

In this sense, the only other ones that would be regex based is the srcset and google optimize one.

charlespwd:
Also a common thing that happens is this:
```liquid
<script src="{{ 'foo.js' | asset_url }}" async></script>
```

Which doesn't work as an independent fragment.

macournoyer:
For all the variables (`{{ ... }}`), we could just ignore those, the same as if `x = ""` in `{{ x }}`. So would end up parsing:

```liquid
<script src="" async></script>
```

As a fragment.

But, I get it. The solution is slowly turning more complicated than the problem.

Regexp sounds good for now. We can revisit once we have a few more HTML checks in place.

macournoyer (approved):
LETS KILL THOSE BLOCKING SCRIPTS!!1 🔥

charlespwd:
True, true! But then what if "async" _is_ the output of the drop?

CHECKMATE!  

😛 

LET'S GET EM! :fire:

mmorissette:
This is so good 🚀

PR #148, opened 2021-02-10
Refactor InMemory Stuff -> {InMemory,FileSystem}Storage

charlespwd:
Follow up on #145

From
```ruby
class InMemoryTemplate < Template; end
class InMemoryJsonFile < JsonFile; end
class InMemoryTheme < Theme; end
```

to 

```ruby
class Storage; end
class InMemoryStorage < Storage; end
class FileSystemStorage < Storage; end

class Theme
  def initialize(storage); end
end

class Template
  def initialize(name, storage); end
end

class JsonFile
  def initialize(name, storage); end
end
```


lib/theme_check/in_memory_storage.rb
```diff
@@ -0,0 +1,37 @@
+# frozen_string_literal: true
+
+# An in-memory theme is not written to disk. The reasons why you'd
```
charlespwd, line 3: *Storage

test/test_helper.rb
```diff
@@ -20,26 +20,41 @@ def analyze_theme(*check_classes, templates)
     end
 
     def make_theme(files = {})
+      # make_file_system_theme(files)
+      make_in_memory_theme(files)
```
charlespwd, line 24: One of those two is commented out because I was running the tests with both to make sure everything passes regardless of which kind of storage you're using. Would there be a better way to make it so all the make_theme tests run on both?

lib/theme_check/analyzer.rb
```diff
@@ -35,12 +35,6 @@ def analyze_theme
       @offenses
     end
 
-    def analyze_file(path)
```
macournoyer, line 38: Nice cleanup! Wasn't used indeed.

lib/theme_check/corrector.rb
```diff
@@ -9,27 +9,23 @@ def initialize(template:)
     def insert_after(node, content)
       line = @template.full_line(node.line_number)
       line.insert(node.range[1] + 1, content)
-      @template.update!
```
macournoyer, line 12: This was to prevent re-writing all template files that haven't been modified after correction. To prevent touching files needlessly.

lib/theme_check/language_server/handler.rb
```diff
@@ -42,7 +42,11 @@ def on_text_document_did_open(_id, params)
       def analyze_and_send_offenses(file_path)
         root = ThemeCheck::Config.find(file_path) || @root_path
         config = ThemeCheck::Config.from_path(root)
-        theme = ThemeCheck::Theme.new(config.root, ignored_patterns: config.ignored_patterns)
+        storage = ThemeCheck::FileSystemStorage.new(
+          config.root,
+          ignored_patterns: config.ignored_patterns
```
macournoyer, line 47: Nice! This makes a lot more sense here.

lib/theme_check/template.rb
```diff
@@ -3,16 +3,25 @@
...
+      @storage.path(@relative_path)
     end
 
     def relative_path
-      @path.relative_path_from(@root)
+      @relative_pathname ||= Pathname.new(@relative_path)
+    end
+
+    def source
+      @storage.read(@relative_path)
```
macournoyer, line 20: This probably should be memoized like before. Source will not change, and we read the template source from a few different places.

lib/theme_check/template.rb
```diff
@@ -31,15 +40,17 @@ def snippet?
...
-    def source
-      @source ||= @path.read
-    end
-
     def lines
       # Retain trailing newline character
       @lines ||= source.split("\n", -1)
     end
 
+    # Not entirely obvious but lines is mutable, corrections are to be
```
macournoyer, line 48: Instead we should just reset `@lines` in `write`, this way `lines` will always be the correct method to call:

```ruby
def write
  # ...
  @lines = nil
end
```

lib/theme_check/theme.rb
```diff
@@ -4,19 +4,33 @@
...
   class Theme
     DEFAULT_LOCALE_REGEXP = %r{^locales/(.*)\.default$}
-    attr_reader :root
+    LIQUID_REGEX = /\.liquid$/i
+    JSON_REGEX = /\.json$/i
 
-    def initialize(root, ignored_patterns: [])
-      @root = Pathname.new(root)
-      @ignored_patterns = ignored_patterns
+    # temp
```
macournoyer, line 10: Why temp?

test/test_helper.rb
macournoyer: The only way I can think of is to use an env var: `TEST_STORAGE=file_system bundle exec rake`, then check `ENV["TEST_STORAGE"]` here.

You could create a rake task to run each one, and run both by default.

test/theme_test.rb
```diff
@@ -52,7 +52,7 @@ def test_templates
 
   def test_snippets
     @themes.each do |theme|
-      assert_equal(["snippets/product"], theme.snippets.map(&:name))
+      assert_equal(["snippets/product"], theme.snippets.map(&:name), theme.storage.class.name)
```
macournoyer, line 55: The last arg of `assert_equal` is the error message. Is this what you want? Might be leftover debug code?

macournoyer (approved):
Beautiful refactoring 👏 

Just a few optimization ideas.

lib/theme_check/template.rb
charlespwd: Indeed!

lib/theme_check/corrector.rb
charlespwd: I thought I put a check in FileSystemStorage that prevents against this in the write function.

But I forgot :upside_down_smile:

My initial idea was to prevent against this by comparing updated_source with source before writing. Must have lost it in the intermediate refactor.

lib/theme_check/theme.rb
charlespwd: If I recall correctly, I only wanted to expose this while unit testing and shouldn't be part of the public API. I left the note there to remind me to remove it :p Clearly I didn't remember :)

lib/theme_check/theme.rb
```diff
@@ -4,19 +4,33 @@
...
+    attr_reader :storage
+
+    def initialize(storage)
+      @storage = storage
     end
 
+    # Expects to return an array of Template
```
charlespwd, line 17: Comment no longer relevant (made sense when NotImplementedError was raised)

test/test_helper.rb
charlespwd: Why, of course! Nice!

test/theme_test.rb
charlespwd: Yeah. Left over debug code to tell me which storage implementation was causing issues.

lib/theme_check/template.rb
charlespwd: Not sure I understand this one. The intention behind this method was to do a check on source != updated_source before writing. And to use it in tests rather than the path.read we had in test_helper.

lib/theme_check/corrector.rb
charlespwd: Just changed it here
https://github.com/Shopify/theme-check/pull/148/files#diff-bcabb8a994ed8297a048e60941f0496b0d45c2d209aa04e995276868241d0713R23-R29

lib/theme_check/storage.rb
```diff
@@ -0,0 +1,21 @@
...
+
+    def write(relative_path, content)
+      raise NotImplementedError
+    end
+
+    def path(relative_path)
+      raise NotImplementedError
+    end
+
+    def files
```
charlespwd, line 17: Missing directories

test/template_test.rb
```diff
@@ -9,38 +9,30 @@ def setup
...
-        content
-      ),
-    ]
+    files = {
+      "templates/index.liquid" => content,
+    }
+    storage = make_storage(files)
+    @template = ThemeCheck::Template.new(
+      "templates/index.liquid",
+      storage
```
charlespwd, line 18: Now that we're back at one. Can probably inline a couple of those vars.

PR #149, opened 2021-02-11
Enable ParserBlockingJavaScript rule by default and fixup

charlespwd:
In this PR:

* Turn the rule on by default
* Allow inline script tags

### DIFF
--- a/lib/theme_check/checks/parser_blocking_javascript.rb
+++ b/lib/theme_check/checks/parser_blocking_javascript.rb
@@ -5,7 +5,12 @@ module ThemeCheck
     severity :error
     category :liquid
 
-    PARSER_BLOCKING_SCRIPT_TAG = /<script(?:(?!defer|async|type=["']module['"]).)*?>/im
+    PARSER_BLOCKING_SCRIPT_TAG = %r{
+      <script                                    # Find the start of a script tag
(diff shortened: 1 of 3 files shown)


lib/theme_check/checks/parser_blocking_javascript.rb
```diff
@@ -5,7 +5,11 @@ class ParserBlockingJavaScript < LiquidCheck
     severity :error
     category :liquid
 
-    PARSER_BLOCKING_SCRIPT_TAG = /<script(?:(?!defer|async|type=["']module['"]).)*?>/im
+    # How to read the following regex:
```
macournoyer, line 8: You can use the `x` option to enable extended regexp syntax in Ruby (ignores whitespaces, so can split on multiple lines & indent), eg.:

```ruby
REGEX = %r{
  <style
  |
  display:\s*(block|inline|flex)
  |
  color:\s*\#\h+
  |
  [\w\-]+:\s*\d+(px|em)
}x
```

Then put the comments inline

macournoyer (approved):

lib/theme_check/checks/parser_blocking_javascript.rb
charlespwd: So much learning :) TY

PR #151, opened 2021-02-12
Bump version & add release doc

macournoyer:


saulsolorzano:
Hi! When can we expect this to be released? The docs are already updated with the new `exclude_schema` parameter in the `TemplateLength` check but the current version 0.2.2 does not support it yet.

Thanks!!

PR #160, opened 2021-02-17
Ignore snippets in UndefinedObject

### LINKED ISSUE #159: Disable running UndefinedObject in snippets
The optional parameters make this impossible.

Either we add an option to only disable it in snippets, or disable the check entirely.

Context: https://shopify.slack.com/archives/C01F4ASU4G7/p1613572406096700?thread_ts=1613513255.092400&cid=C01F4ASU4G7

macournoyer:
See #134 for a previous attempt at a fix. But we ended up recommending to use `default` on all optional parameters. But this is not a good solution.

See context in: https://shopify.slack.com/archives/C01F4ASU4G7/p1613572406096700?thread_ts=1613513255.092400&cid=C01F4ASU4G7

Can be turned back on w/ `ignore_snippets: false`

Fixes #159

### DIFF
--- a/config/default.yml
+++ b/config/default.yml
@@ -67,7 +67,7 @@ MissingRequiredTemplateFiles:
 
 UndefinedObject:
   enabled: true
-  ignore_snippets: true
+  exclude_snippets: true
 
 RequiredDirectories:
   enabled: true
--- a/config/default.yml
+++ b/config/default.yml
@@ -67,6 +67,7 @@ MissingRequiredTemplateFiles:
 
 UndefinedObject:
   enabled: true
+  ignore_snippets: true
 
 RequiredDirectories:
   enabled: true
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -7,6 +7,7 @@ AllCops:
   TargetRubyVersion: 2.7
   Exclude:
     - 'vendor/bundle/**/*'
+    - 'packaging/builds/**/*'
 
 Metrics/MethodLength:
   Enabled: false
--- a/lib/theme_check/checks/undefined_object.rb
+++ b/lib/theme_check/checks/undefined_object.rb
@@ -49,23 +49,28 @@ module ThemeCheck
       end
     end
 
-    def initialize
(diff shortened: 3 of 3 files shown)


config/default.yml
```diff
@@ -67,6 +67,7 @@ MissingRequiredTemplateFiles:
 
 UndefinedObject:
   enabled: true
+  ignore_snippets: true
```
charlespwd, line 70: Should this be the opposite? default to false and can be turned on?

macournoyer: `dont_ignore_snippets: false`? :p

Good point! renaming

macournoyer: Actually, would it make more sense to name it `exclude_snippets: true` ?

To mimic the other one we already have: https://github.com/Shopify/theme-check/blob/master/config/default.yml#L33

charlespwd: Actually I'm saying this should be opt-in in Dawn rather than by default to everyone.

macournoyer: Oh I don't think this will roll. We want to use the same things internally that we'll promote externally.

So maybe this shouldn't even be an option? Always ignore the snippets.

charlespwd: Right but what's the difference between a typo and an optional parameter in this case? Should you be flagged if you use a typo'ed argument?

macournoyer: Yes I agree that we should. But I can't think of a way to allow optional parameters in snippets, and check for this at the same time.

I think we had two choices:
1. Force users to declare optional parameters (using `| default: ...`)
2. Stop checking snippets for this, because snippets are not

I wished we could stick w/ 1, but I agree it does make the code more complicated, as mentioned Slack.

And we're going w/ 2, because "we don't want to encourage the use of snippets as functions" anyways.

Maybe there's another solution, but I don't see it atm.

charlespwd: Alright alright :p

charlespwd (approved):

jplhomer (approved):
❤️ I echo all of the smart things you and @charlespwd have discussed 😛

PR #161, opened 2021-02-17
Add Completion Engine

charlespwd:
Adds a completion engine to the Language Server!

🎥 👉 [Video](https://screenshot.click/25-42-7yvzq-tu36j.mp4)

## High level view of how it works:

The language client sends updated buffers to the language server on the following events:
- textDocument/open
- textDocument/change
- textDocument/close

Those buffers are persisted inside an InMemoryStorage instance.

When the language client sends a completion request, we do the following:

1. Extract the file URI and cursor position (line, character) from the request params
2. Split the buffer on tag/variable start and end (with lookahead/behinds). 
3. Grab all completions for the current token and cursor position
4. Send them back to the language client

## How completions are setup:

We have two main concepts: 

1. Completion Engine - responsible for bridging between the InMemoryStorage, the Completion Providers and the language client
2. Completion Providers - Given a token and a cursor position, outputs a list of completions. An empty array if the current position or token cannot be completed. 

The architecture is very similar to how checks are implemented. We have a parent CompletionProvider class that defines the API (and tracks the subclasses in an array), and the individual completion providers (tag, object, filter, ... extensible) implement that class. The CompletionEngine instantiates one of each and flat_maps the result of calling completions on every single one. 

```
language_server/
├── ...
├── completion_engine.rb
├── completion_helper.rb
├── completion_provider.rb
└── completion_providers/
    ├── filter_completion_provider.rb
    ├── object_completion_provider.rb
    └── tag_completion_provider.rb
```

## For future PRs:
- [ ] Complete known object properties inside drops - `{{ obj.<property>`
- [ ] Complete variables inside tags
- [ ] Complete arguments inside filters
- [ ] Complete known object properties inside assignments, if statements, for loops, etc.


lib/theme_check/language_server/handler.rb
```diff
@@ -31,14 +37,52 @@ def on_initialize(id, params)
...
 
+      def text_document_uri(params)
+        params.dig('textDocument', 'uri').sub('file://', '')
+      end
+
+      def did_change_text(params)
+        params.dig('contentChanges', 0, 'text')
+      end
+
+      def template(params)
```
charlespwd, line 82: This is no longer a template, it should be `text`

lib/theme_check/language_server/completion_engine.rb
```diff
@@ -0,0 +1,45 @@
...
+        end
+      end
+
+      private
+
+      def cursor_index(token, line, col)
+        relative_line = line - token.start_line
+        return col - token.start_col if relative_line == 0
+        from_line_column_to_index(relative_line, col, token.content)
+      end
```
charlespwd, line 42: I might be able to simplify this even more by converting to index (instead of line/col) earlier.

test/language_server/completion_engine_test.rb
```diff
@@ -0,0 +1,101 @@
+# frozen_string_literal: true
+require "test_helper"
+
+module ThemeCheck
+  module LanguageServer
+    class CompletionEngineTest < Minitest::Test
```
macournoyer, line 6: Not sure if it's relevant. Should there be some tests w/ missing closing tokens at this level? Missing: `%}`, `}}` ? Worth it? I see it is tested bellow.

macournoyer (approved):
God damn this it awesome! 👏  🔥  Amazing work!

test/language_server/completion_engine_test.rb
charlespwd: Definitely, since this is the kind of thing that bit me during the demo video :p To be fixed soon :)

data/shopify_liquid/tags.yml
```diff
@@ -0,0 +1,26 @@
+---
```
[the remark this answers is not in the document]
charlespwd: I feel like we'd be missing a couple ones that are Shopify specific. e.g. `stylesheet`, `schema`, etc.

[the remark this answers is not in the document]
macournoyer: All [here](https://github.com/Shopify/theme-check/blob/master/lib/theme_check/tags.rb). The parsing code from those tags is needed to parse the templates.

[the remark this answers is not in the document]
charlespwd: Actually it looks like we're missing some:
```
irb(main):003:0> Liquid::Template.tags.each.map {|k, _| k}.size
=> 18
```
And this file has 26

PR #174, opened 2021-02-22
Add init command to CLI

### LINKED ISSUE #100: Add flag for generating .theme-check.yml
Some other linters such as ESLint allow users to pass an `--init` flag to generate a config (documented [here](https://eslint.org/docs/user-guide/getting-started#installation-and-usage)) . This flag could probably just copy the `config/default.yml`. It would be useful for people who maintain multiple themes to be able to use this.

yishus:
Resolves #100

Running theme-check with the `--init` flag will not run the checks, instead it will check if a `.theme-check.yml` config file is present. If it is it will abort and print that the file exists, if not it will create a new file that is cloned from `config/default.yml`.


macournoyer (approved):
Awesome! 👏

PR #192, opened 2021-03-09
Fix multiple occurrences of UndefinedObject not being reported

### LINKED ISSUE #186: Problems detecting the same type error multiple times in the file
When particular instance of a particular type of error occurs, it seems theme-check can only report/detect it once in the same file:

image

Expected behaviour would be that each error instance can be reported separately.

macournoyer:
Fixes #186

The issue seemed to be caused by missing line_number for VariableLookup nodes.

Solution: Walk back the parent nodes until we find a line number.


charlespwd (approved):
LGTM!

Issue #189, opened 2021-03-08
Implement Size Limit Checks


charlespwd:
This would be a check of type :performance.

For server rendered pages, it is an anti-pattern to execute large JavaScript bundles on every navigated page. However, that doesn't mean they don't have a reason to exist. Arguably, chat widgets are mini apps embedded inside web pages. Designing such an app with server rendered updates would be absurd.

On the flip side, if 10% of the users interact with the chat widget, the other 90% should not have to execute the entire bundle on every page load.

The natural solution to this problem is to implement the chat widget using the [import on interaction pattern](https://addyosmani.com/blog/import-on-interaction).

So, how can we implement a rule that encourages the use of this pattern while still enabling large bundles inside a theme?

Some ideas:
* ~Static size checks~ _Can't tell if loaded on interaction or not. Prevents the accepted use of large JS bundles loaded on interaction_.
* Static resource import check / aka find all script tags on the page, find their brotli size throw above configurable threshold. To be explored. What about remote assets? How can we figure out if the asset is being used? Scanning for `asset_url` filters?

macournoyer:
Just an idea: We could fetch & discard the remote resources just to get their sizes (if you set `Accept-Encoding: br` is will take care of the compression)? And cache it locally perhaps?

PR #194, opened 2021-03-09
Add AssetSizeJavascript Check

charlespwd:
Related #189 (part 1: JS)

Add an AssetSizeJavaScript check + docs. The goal being to discourage
the use of frameworks and encourage web native.

Excerpt from the checks' docs (general template adapted from ESLint):

> # Prevent JavaScript Abuse on Server Rendered Themes (AssetSizeJavaScript)
>
> For server rendered pages, it is an anti-pattern to execute large JavaScript bundles on every navigation.
>
> This doesn't mean they don't have a reason to exist. For instance, chat widgets are mini applications embedded inside web pages. Designing such an app with server rendered updates would be absurd. However, if only 10% of the users interact with the chat widget, the other 90% should not have to execute the entire bundle on every page load.
>
> The natural solution to this problem is to implement the chat widget using the [Import on Interaction Pattern][ioip].
>
> ## Rule Details
>
> This rule disallows the use of theme JavaScript files and external scripts to have a compressed size greater than a configured `threshold_in_bytes`.
>
> **:-1: Example of incorrect code for this rule:**
> ```liquid
> 
> <script src="{{ 'chat-widget.js' | asset_url }}" defer></script>
>
> 
> <script src="https://code.jquery.com/jquery-3.6.0.min.js" defer></script>
> ```
>
> **:+1: Example of correct code for this rule:**
> ```liquid
> <script>
>   const chatWidgetButton = document.getElementById('#chat-widget')
>
>   chatWidgetButton.addEventListener('click', e => {
>     e.preventDefault();
>     import("{{ 'chat-widget.js' | asset_url }}")
>       .then(module => module.default)
>       .then(ChatWidget => ChatWidget.init())
>       .catch(err => {
>         console.error(err);
>       });
>   });
> </script>
> ```

See docs/checks/asset_size_javascript.md for more details.


lib/theme_check/remote_asset_file.rb
```diff
@@ -0,0 +1,36 @@
...
+        URI.parse(src.sub(%r{^//}, "https://"))
+      end
+    end
+
+    def initialize(src)
+      @uri = RemoteAssetFile.uri(src)
+    end
+
+    def content
+      @content ||= Net::HTTP.get(@uri)
```
macournoyer, line 29: I think this means we'll fetch every remote access each time we analyze the theme?

And we analyze the theme each time a file is saved in LSP.

This looks like a problem. Maybe we should cache this info in a local file?

macournoyer (commented):
Some early comments/questions

lib/theme_check/remote_asset_file.rb
```diff
@@ -0,0 +1,36 @@
...
+    end
+
+    def gzipped_size
+      @gzipped_size ||= Zlib.gzip(content).size
```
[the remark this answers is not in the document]
charlespwd: I don't think we do. I used it this way for consistency with AssetFile (also, what if the server doesn't gzip?)

config/default.yml
```diff
@@ -80,3 +80,6 @@ MissingEnableComment:
 
 ParserBlockingJavaScript:
   enabled: true
+
+AssetSizeJavaScript:
+  enabled: true
```
charlespwd, line 85: Might not want enabled by default so we can push this without forcing the 10KB rule just now.

lib/theme_check/remote_asset_file.rb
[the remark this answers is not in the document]
macournoyer: If we set the same header as Chrome (`Accept-Encoding: gzip, deflate, br`), the server will just pick the best one. It will be the same bytes as the browser sees it, which it what you want to measure.

lib/theme_check/remote_asset_file.rb
macournoyer: Aaaah I missed that. Excellent!

lib/theme_check/regex_helpers.rb
```diff
@@ -0,0 +1,15 @@
+# frozen_string_literal: true
+
+module ThemeCheck
+  module RegexHelpers
+    def matches(s, re)
```
macournoyer, line 5: IIUC what this code does, you're looking for `String#scan`: https://ruby-doc.org/core-3.0.0/String.html#method-i-scan

lib/theme_check/remote_asset_file.rb
```diff
@@ -0,0 +1,45 @@
+# frozen_string_literal: true
+require "net/http"
+require "pathname"
+require "zlib"
```
macournoyer, line 4: Not used anymore

lib/theme_check/remote_asset_file.rb
```diff
@@ -0,0 +1,45 @@
...
+        req = Net::HTTP::Get.new(@uri)
+        req['Accept-Encoding'] = 'gzip, deflate, br'
+        http.request(req)
+      end
+
+      @content = res.body
+    end
+
+    def gzipped_size
+      @gzipped_size ||= @content.size
```
macournoyer, line 42: `bytesize` would be more accurate, especially for encoded content: https://ruby-doc.org/core-3.0.0/String.html#method-i-bytesize

config/default.yml
```diff
@@ -80,3 +80,6 @@ MissingEnableComment:
 
 ParserBlockingJavaScript:
   enabled: true
+
+AssetSizeJavaScript:
+  enabled: false
```
macournoyer, line 85: Should add the default for `threshold_in_bytes` here too:

```suggestion
  enabled: false
  threshold_in_bytes: 10000
```

macournoyer (approved):
Straight fire 🔥

lib/theme_check/checks/asset_size_javascript.rb
```diff
@@ -0,0 +1,76 @@
...
+        [^>]*                              # any non closing character till the end
+      >
+    }omix
+
+    attr_reader :threshold_in_bytes
+
+    def initialize(threshold_in_bytes: 10000)
+      @threshold_in_bytes = threshold_in_bytes
+      @files = []
+      @scripts = []
```
charlespwd, line 34: Don't need either of these anymore.

lib/theme_check/regex_helpers.rb
macournoyer: You can use `https://ruby-doc.org/core-3.0.0/Regexp.html#method-c-last_match`

Issue #195, opened 2021-03-10
Improve documentation of checks and add links to documentation from within the editor.


charlespwd:
ESLint et al. offer links to rule definitions and options to browse to them from within the editor:

* [Docs](https://eslint.org/docs/developer-guide/working-with-rules)
* [Example rule docs](https://eslint.org/docs/rules/eqeqeq)

![](https://screenshot.click/10-59-8tpvd-3ns74.png)

We should do the same. We could start by copying that template and include our docs in the docs/checks folder.

charlespwd:
I thought the only thing I needed to do was add the codeDescription attribute on diagnostics but it looks like it's not behaving the same way as ESLint.

I also noticed that they have code actions for showing documentation. Might want to add that.

charlespwd:
Stacks of context:
- https://github.com/lifeart/ember-language-server/pull/171

charlespwd:
Took me a while to figure this out, but the codeDescriptionSupport client capability is only supported by `vscode-languageclient` v7.0.0+. We'll need to update that in [theme-check-vscode](https://github.com/Shopify/theme-check-vscode).

PR #196, opened 2021-03-10
Add Documentation Template + TemplateLength Check Documentation

charlespwd:
Add a check documentation template + a way to link to the documentation (from within the repo) in check classes.

This also implements the codeDescription property in diagnostics so that the docs can be accessed from within the editor:

https://screenshot.click/10-29-cjx7r-4asor.mp4

Related #195

### DIFF
--- a/docs/checks/template_length.md
+++ b/docs/checks/template_length.md
@@ -18,7 +18,7 @@ This check is aimed at eliminating large template files.
 
 The default configuration for this check is the following:
 
-```
+```yaml
 TemplateLength:
   enabled: true
   max_length: 200
--- a/docs/checks/template_length.md
+++ b/docs/checks/template_length.md
@@ -1,0 +1,50 @@
+# Discourage the use of large template files (`TemplateLength`)
+
+This check exists to discourage the use of large template files. Use snippets to componentize your theme.
+
+## Check Details
+
+This check is aimed at eliminating large template files.
+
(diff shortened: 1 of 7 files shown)


docs/checks/template_length.md
```diff
@@ -0,0 +1,50 @@
...
+
+:+1: Examples of **correct** code for this check:
+
+- Files that have less lines than the threshold
+
+## Check Options
+
+The default configuration for this check is the following:
+
+```
```
charlespwd, line 21: yaml.

lib/theme_check/checks/template_length.rb
```diff
@@ -3,6 +3,7 @@ module ThemeCheck
   class TemplateLength < LiquidCheck
     severity :suggestion
     category :liquid
+    doc docs_url("docs/checks/template_length.md")
```
macournoyer, line 6: Since this will only be used in checks, should the `docs/checks/` prefix be moved to the helper method?

macournoyer (approved):

lib/theme_check/checks/template_length.rb
charlespwd: But at the same time that's very magicky. Doesn't make it clear where the docs are really. Whereas with the full path here you can pretty much copy paste a Check without context and understand where the docs are supposed to be.

PR #200, opened 2021-03-11
Documentation Spree (Part deux)

charlespwd:
Add documentation for the following checks:

  * SpaceInsideBraces
  * RequiredLayoutThemeObject
  * RequiredDirectories
  * ParserBlockingJavaScript
  * NestedSnippet
  * MissingTemplate
  * MissingRequiredTemplateFiles

### DIFF
--- a/docs/checks/matching_schema_translations.md
+++ b/docs/checks/matching_schema_translations.md
@@ -71,6 +71,15 @@ Add checks for eliminating translations mistakes in schema tags.
 {% endschema %}
 ```
 
+## Check Options
+
+The default configuration for this check is the following:
+
+```yaml
+MatchingSchemaTranslations:
+  enabled: true
+```
+
 ## Version
 
 This check has been introduced in Theme Check 0.1.0.
--- a/docs/checks/nested_snippet.md
+++ b/docs/checks/nested_snippet.md
@@ -4,7 +4,7 @@ Reports deeply nested `render` tags (or deprecated `include` tags).
 
 ## Check Details
 
-This check is aimed at eliminating ...
+This check is aimed at eliminating excessive nesting of snippets.
 
 :-1: Examples of **incorrect** code for this check:
 
--- a/docs/checks/nested_snippet.md
+++ b/docs/checks/nested_snippet.md
@@ -1,0 +1,69 @@
+# Prevent deeply nested snippets (`NestedSnippet`)
+
+Reports deeply nested `render` tags (or deprecated `include` tags).
+
+## Check Details
+
+This check is aimed at eliminating ...
+
+:-1: Examples of **incorrect** code for this check:
+
+```liquid
+{% comment %}templates/index.liquid{% endcomment %}
+  {% render 'one' %}
+
+{% comment %}snippets/one.liquid{% endcomment %}
+  {% render 'two' %}
+
+{% comment %}snippets/two.liquid{% endcomment %}
+  {% render 'three' %}
+
+{% comment %}snippets/three.liquid{% endcomment %}
+  {% render 'four' %}
+
+{% comment %}snippets/four.liquid{% endcomment %}
+  ok
+```
+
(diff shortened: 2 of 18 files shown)


docs/checks/nested_snippet.md
```diff
@@ -0,0 +1,69 @@
+# Prevent deeply nested snippets (`NestedSnippet`)
+
+Reports deeply nested `render` tags (or deprecated `include` tags).
+
+## Check Details
+
+This check is aimed at eliminating ...
```
charlespwd, line 7: Complete ME.

macournoyer: excessive nesting of snippets

docs/checks/matching_schema_translations.md
```diff
@@ -71,6 +71,15 @@ Add checks for eliminating translations mistakes in schema tags.
 {% endschema %}
 ```
 
+## Check Options
+
+
+```yaml
+MatchingSchemaTranslations:
```
macournoyer, line 79: I'm a bit sad that now we got this duplicated in 3 places:

- `config/default.yml`
- the check's `initialize`
- the docs

I don't have a solution. But something to think about...

But we must update the CONTRIBUTING.md file to document this at least.

macournoyer: Just an idea, if we generated the (allow some ERB in the docs) we could dynamically generate this part from the config/default.yml file.

lib/theme_check/checks/missing_required_template_files.rb
```diff
@@ -3,11 +3,11 @@
 module ThemeCheck
   # Reports missing shopify required theme files
   # required templates: https://shopify.dev/tutorials/review-theme-store-requirements-files
-
+  # https://shopify.dev/docs/themes/theme-templates
```
macournoyer, line 6: Is this needed?

lib/theme_check/checks/matching_schema_translations.rb
```diff
@@ -3,7 +3,7 @@ module ThemeCheck
   class MatchingSchemaTranslations < LiquidCheck
     severity :suggestion
     category :translation
-    doc docs_url("checks/docs/matching_schema_translations.rb")
+    doc docs_url("docs/checks/matching_schema_translations.md")
```
macournoyer, line 6: The `doc` could default to `docs/checks/[NAME OF CHECK].md` (if the file exists), so you wouldn't even need to specify here.

macournoyer (approved):
A few suggestions.

Also disclaimer that I'm notoriously bad at reviewing docs. Perhaps we could get a tech writer to review those?

lib/theme_check/checks/missing_required_template_files.rb
charlespwd: I moved it from below and also in the documentation file. Can definitely remove.

lib/theme_check/checks/matching_schema_translations.rb
charlespwd: Yeah. My initial thoughts were that this was more explicit. So anybody that opens the project for the first time suddenly knows where to find the documentation files. Whereas by making it magic you need to learn the magic otherwise you wouldn't know.

But we could make it magic and make a unit test that fails when a doc file doesn't exist for a new check. Plus make a script that generates one for you.

docs/checks/matching_schema_translations.md
charlespwd: I feel you. I don't have a good solution to that either. If we do make a microsite I think this would be a good idea. But since now I'm linking directly to the markdown files in the docs, I don't think we can do this just yet.

lib/theme_check/checks/matching_schema_translations.rb
charlespwd: I'm adding all the docs and will change this after once every check is documented.

PR #206, opened 2021-03-15
Add AssetSizeCSS Check

### LINKED ISSUE #189: Implement Size Limit Checks

charlespwd:
Fixes #189

This check is a lot like AssetSizeJavaScript but for CSS.

- Disabled by default
- Default value = 100KB
- Link to Alex Russells' Performance Inequality Gap in docs

### DIFF
--- a/lib/theme_check/checks/asset_size_css.rb
+++ b/lib/theme_check/checks/asset_size_css.rb
@@ -1,0 +1,95 @@
+# frozen_string_literal: true
+module ThemeCheck
+  class AssetSizeCSS < LiquidCheck
+    include RegexHelpers
+    severity :error
+    category :performance
+    doc docs_url(__FILE__)
+
+    Link = Struct.new(:href, :index)
+
+    TAG = /#{Liquid::TagStart}.*?#{Liquid::TagEnd}/om
+    VARIABLE = /#{Liquid::VariableStart}.*?#{Liquid::VariableEnd}/om
+    START_OR_END_QUOTE = /(^['"])|(['"]$)/
+    LINK_TAG_HREF = %r{
+      <link
+        (?=(?:[^>]|\n|\r)+?rel=['"]?stylesheet['"]?)+?  # Make sure rel=stylesheet is in the link with lookahead
+        [^>]+                                           # any non closing tag character
+        href=                                           # href attribute start
+        (?<href>
+          '(?:#{TAG}|#{VARIABLE}|[^']+)*'|              # any combination of tag/variable or non straight quote inside straight quotes
+          "(?:#{TAG}|#{VARIABLE}|[^"]+)*"               # any combination of tag/variable or non double quotes inside double quotes
+        )
+        [^>]*                                           # any non closing character till the end
+      >
+    }omix
+    STYLESHEET_TAG = %r{
+      #{Liquid::VariableStart}          # VariableStart
+      (?:(?!#{Liquid::VariableEnd}).)*? # anything that isn't followed by a VariableEnd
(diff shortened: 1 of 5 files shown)


lib/theme_check/checks/asset_size_css.rb
```diff
@@ -0,0 +1,98 @@
...
+    end
+
+    def record_offenses
+      stylesheets(@source).each do |stylesheet|
+        file_size = href_to_file_size(stylesheet.href)
+        next if file_size.nil?
+        next if file_size <= threshold_in_bytes
+        add_offense(
+          "CSS on every page load exceding compressed size threshold (#{threshold_in_bytes} Bytes).",
+          node: @node,
```
macournoyer, line 56: If this offense is template-wide, you should pass `template: node.template` instead.

test/config_default_test.rb
```diff
@@ -0,0 +1,26 @@
...
+require "test_helper"
+require "yaml"
+
+module ThemeCheck
+  class ConfigDefaultTest < Minitest::Test
+    def setup
+      @default_config = YAML.load(default_config_path.read)
+    end
+
+    def test_all_checks_are_in_config_default_yml
```
macournoyer, line 11: Brilliant!

lib/theme_check/checks/asset_size_css.rb
```diff
@@ -0,0 +1,98 @@
+# frozen_string_literal: true
+module ThemeCheck
+  # Reports errors when trying to use too much JavaScript on page load
```
macournoyer, line 3: Oops! 
```suggestion
  # Reports errors when trying to use too much CSS on page load
```

Maybe the comment is not needed now that we got some awesome doc?

macournoyer (approved):

lib/theme_check/checks/asset_size_css.rb
charlespwd: It isn't. It highlights the URL (or liquid variable) of imported CSS files that are too big.

lib/theme_check/checks/asset_size_css.rb
charlespwd: True that :p

charlespwd: Copy paste forgot to change the comment. :sweat_smile:

lib/theme_check/checks/asset_size_css.rb
macournoyer: But since you using the `@node` from the `on_document`, the offense will always point the whole document.

That's the top level node of the Liquid AST.

PR #212, opened 2021-03-16
Do not flag SystemTranslations as errors

### LINKED ISSUE #114: Don't report Shopify-defined translation keys as missing
This tool is *amazing*.

Small thing - this error is an edge case and shouldn't come up because the string is defined by Shopify, not the theme. Shopify detects if it exists before deciding on showing the Google Captcha button on the lower right of the screen 

image

charlespwd:
Fixes #114

This comes with the following asterisk:

- Only translations surfaced in the LanguageEditor are considered OK.


lib/theme_check/checks/translation_key_exists.rb
```diff
@@ -1,5 +1,20 @@
 # frozen_string_literal: true
 module ThemeCheck
+  module SystemTranslations
```
macournoyer, line 3: Why not put this in `lib/theme_check/shopify_liquid` with the other similar files?

lib/theme_check/checks/translation_key_exists.rb
```diff
@@ -23,8 +38,9 @@ def on_variable(node)
 
     private
 
-    def key_exists?(key)
-      pointer = @theme.default_locale_json.content
+    def key_exists?(key, hash)
```
macournoyer, line 41: Maybe a name to clarify its use:

```suggestion
    def key_exists?(key, default_locale)
```

charlespwd: Yeah, we don't need this anymore. I did initially when the file was also a hash. Opted to go with an array of keys instead (this way we only need the keys and not the content). 

I'll remove this change.

lib/theme_check/checks/translation_key_exists.rb
charlespwd: Ah... I remember why. Because shopify_liquid is for liquid features. Objects, filters, etc. This kind of doesn't fit that mold. This is also why the data file is at the top of /data and not in /data/shopify_liquid. 

I definitely could extract this into its own file though. But it's not used anywhere else and would require changes to theme_check.rb as well to require it.

macournoyer: Got it! Maybe `data/shopify/liquid` & `data/shopify/locales` would make sense?

macournoyer (approved):

lib/theme_check/checks/translation_key_exists.rb
charlespwd: data/shopify/liquid kind of does but data/shopify/locales not really since I'm not importing the translation strings themselves, only the keys. So it'd be

```
data/
  shopify/
     liquid/
        ...
     system_translations.yml
lib/
  theme_check/
    shopify/
       liquid/
          ...
       system_translations
    ...
```


Whereas now it's
```
data/
  shopify_liquid/
    ...
  system_translation_keys.yml
lib/
  theme_check/
    shopify_liquid/
       ...
    checks/
       translation_key_exists.rb
```

charlespwd: IDK feels a bit early to extract and move all those files.

PR #223, opened 2021-03-23
Add Snippet Completion + Document Links

### LINKED ISSUE #180: Complete render strings
https://shopify.dev/docs/themes/liquid/reference/tags/theme-tags#render 

The `{% render '<file>' %}` tag renders files from the snippets folder. We could pretty easily add a RenderSnippetCompletionProvider that gets a name of those files and offers them as suggestions.

charlespwd:
Fixes #180 

In this PR:

* Add `{% render '..' %}` completions
* Make snippets in `{% render '..' %}` clickable links

Demo: https://screenshot.click/23-09-71h84-pp23z.mp4
Also: https://screenshot.click/23-22-5tyee-kv5vl.mp4

### DIFF
--- a/lib/theme_check/in_memory_storage.rb
+++ b/lib/theme_check/in_memory_storage.rb
@@ -6,20 +6,28 @@
 # as a big hash already, leave it like that and save yourself some IO.
 module ThemeCheck
   class InMemoryStorage < Storage
-    def initialize(files = {})
+    def initialize(files = {}, root = nil)
       @files = files
+      @root = root
     end
 
     def path(name)
+      return File.join(@root, name) unless @root.nil?
+      name
+    end
+
+    def relative_path(name)
+      path = Pathname.new(name)
+      return path.relative_path_from(@root).to_s unless path.relative? || @root.nil?
       name
     end
 
     def read(name)
-      @files[name]
+      @files[relative_path(name)]
     end
 
     def write(name, content)
-      @files[name] = content
+      @files[relative_path(name)] = content
     end
 
     def files
--- a/lib/theme_check/language_server.rb
+++ b/lib/theme_check/language_server.rb
@@ -1,5 +1,6 @@
 # frozen_string_literal: true
 require_relative "language_server/protocol"
+require_relative "language_server/constants"
 require_relative "language_server/handler"
 require_relative "language_server/server"
 require_relative "language_server/tokens"
@@ -7,6 +8,7 @@ require_relative "language_server/position_helper"
 require_relative "language_server/completion_helper"
 require_relative "language_server/completion_provider"
 require_relative "language_server/completion_engine"
+require_relative "language_server/document_link_engine"
 
 Dir[__dir__ + "/language_server/completion_providers/*.rb"].each do |file|
   require file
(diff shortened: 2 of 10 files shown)


lib/theme_check/language_server.rb
```diff
@@ -1,12 +1,14 @@
...
 require_relative "language_server/protocol"
+require_relative "language_server/constants"
 require_relative "language_server/handler"
 require_relative "language_server/server"
 require_relative "language_server/tokens"
 require_relative "language_server/position_helper"
 require_relative "language_server/completion_helper"
 require_relative "language_server/completion_provider"
 require_relative "language_server/completion_engine"
+require_relative "language_server/document_links_engine"
```
charlespwd, line 11: Maybe should singularize this.

lib/theme_check/language_server/document_link_engine.rb
```diff
@@ -0,0 +1,47 @@
...
+      include PositionHelper
+      include RegexHelpers
+
+      def initialize(storage)
+        @storage = storage
+      end
+
+      def document_links(uri)
+        buffer = @storage.read(uri)
+        matches(buffer, PARTIAL_RENDER).map do |match|
```
macournoyer, line 15: Is this file specific to `render` support?

lib/theme_check/language_server/handler.rb
```diff
@@ -71,6 +81,23 @@ def on_text_document_completion(id, params)
...
+
+        # Make a real FS to get the files from the snippets folder
+        fs = ThemeCheck::FileSystemStorage.new(
+          config.root,
+          ignored_patterns: config.ignored_patterns
+        )
+
+        # Turn that into a hash of empty buffers
+        files = fs.files
+          .map { |fn| [fn, ""] }
```
macournoyer, line 95: Just curious why you need to do this?

charlespwd: Because I need a list of snippets from the RenderSnippetCompletionProvider. The InMemoryStorage doesn't have anything in it and I didn't want to have both an InMemoryStorage instance for buffers and a FSStorage for files _just_ so that the render snippet completion provider was able to list those files.

jplhomer (approved):
😍

lib/theme_check/language_server/document_link_engine.rb
charlespwd: For now, yes. But we might want to add support for things that get piped to asset_url in the future.

lib/theme_check/in_memory_storage.rb
```diff
@@ -6,20 +6,28 @@
...
+
+    def relative_path(name)
+      path = Pathname.new(name)
+      return path.relative_path_from(@root).to_s unless path.relative? || @root.nil?
       name
     end
 
     def read(name)
-      @files[name]
+      @files[relative_path(name)]
```
charlespwd, line 26: The reason I'm doing this is because I would otherwise have absolute paths in the InMemoryStorage. So we'd have duplicate entries for absolute paths and relative paths. I could also revert this and only have absolute paths.

lib/theme_check/language_server/handler.rb
macournoyer: Oh I see! Because that's what `InMemoryStorage.new` expects! 👍

macournoyer (approved):
🧨

PR #227, opened 2021-03-23
Make releasing slightly easier with rake prerelease[$version]

charlespwd:


RELEASING.md
```diff
@@ -2,11 +2,18 @@
 
 1. Check the Semantic Versioning page for info on how to version the new release: http://semver.org
 
-2. Create a PR to update the version in `lib/theme_check/version.rb` and replace the `THEME_CHECK_VERSION` placeholder in the documentation for new rules.
+2. Run the following command to update the version in `lib/theme_check/version.rb` and replace the `THEME_CHECK_VERSION` placeholder in the documentation for new rules:
 
-3. Merge your PR to master
+   ```bash
+   VERSION="X.X.X"
+   rake prerelease[$VERSION]
```
macournoyer, line 9: Just curious, why not do it on a single line here?

macournoyer (approved):

RELEASING.md
charlespwd: I mean, you can... I usually do it like that in docs to illustrate what are variables while still making the script executable. You can copy paste this and it would run.   

Whereas if I were to do something like `rake prerelease[X.X.X]` it wouldn't be as obvious.

PR #229, opened 2021-03-24
Add RemoteAsset check

### LINKED ISSUE #214: Explore RemoteAsset Check
Years ago, loading jQuery from a common CDN was a a performance improvement because the browser cache could be reused across website. This is no longer true. Browsers now include the domain from which the resource was loaded from in the cache key.

Therefore, nowadays, this technique makes things worse. And here are a couple of reasons why:

- A new connection dance (DNS, TCP, TLS) must be done to start downloading the resource. This takes 5 round trips to achieve. The farther away the buyer is from that remote domain, the longer it takes. Moreover, browsers cap the number of open connections (IIRC Chrome has a maximum of 6).
- The benefits of HTTP/2 prioritization are lost when assets are delivered from multiple domains. 
- The TCP congestion window scaling must happen on every connection. So you might have connections competing for bandwidth + starting from 0bps instead of 5000mbps.

For a really good talk that dives deep into HTTP/2 and explains some of these concepts, look [here](https://www.youtube.com/watch?v=ct5MvtmL1NM&feature=youtu.be).

With all those things considered, it is best when all assets are served from a small number of domains.

So what we'd need to do is something similar to what we did with `AssetSizeJavaScript` and `AssetSizeCSS` and find resources that are being loaded from external domains and flag them as errors. Encouraging the use of the `assets/` folder so that those files are served by `cdn.shopify.com` instead.

charlespwd:
Fixes #214.

### DIFF
--- a/test/checks/remote_asset_test.rb
+++ b/test/checks/remote_asset_test.rb
@@ -1,0 +1,116 @@
+# frozen_string_literal: true
+require "test_helper"
+
+module ThemeCheck
+  class RemoteAssetTest < Minitest::Test
+    def test_no_offense_for_good_behaviour
+      offenses = analyze_theme(
+        RemoteAsset.new,
+        "templates/index.liquid" => <<~END,
+          <!-- scripts -->
+          <script src="{{ 'theme.js' | asset_url }}" defer></script>
+
+          <!-- styles -->
+          {{ 'theme.css' | asset_url | stylesheet_tag }}
+          <link href="{{ 'theme.css' | asset_url }}" rel="stylesheet">
+
+          <!-- images -->
+          <img alt="logo" src="{{ 'heart.png' | asset_url }}" width="100" height="100">
+          <source src="{{ 'heart.png' | asset_url }}">
+
+          <!-- Good kind of URLs -->
+          <source src="foo.png">
+          <source src="/app/src/foo.png">
+          <source src="blob:...">
+          <source src="data:image/png;base64,...">
+          <source src="{{ image.src | img_url }}">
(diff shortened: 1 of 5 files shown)


jmignac (approved):

test/checks/remote_asset_test.rb
```diff
@@ -0,0 +1,96 @@
...
+        Remote asset should be hosted by Shopify for better performance. at templates/index.liquid:1
+      END
+    end
+
+    def test_flag_use_of_remote_stylesheet
+      offenses = analyze_theme(
+        RemoteAsset.new,
+        "templates/index.liquid" => <<~END,
+          <link href="https://example.com/bootstrap.css" rel="stylesheet">
+          <link href="{{ "https://example.com/bootstrap.css" | replace: 'bootstrap', 'tailwind' }}" rel="stylesheet">
```
macournoyer, line 67: 😂

test/checks/remote_asset_test.rb
```diff
@@ -0,0 +1,96 @@
...
+      END
+    end
+
+    # Note: this highlights how we might have a false positive for assign that uses the asset_url.
+    # This doesn't feel like a common practice though.
+    def test_flag_use_of_html_filter_without_asset_url_filter
+      offenses = analyze_theme(
+        RemoteAsset.new,
+        "templates/index.liquid" => <<~END,
+          {{ url | img_tag }}
```
macournoyer, line 84: This looks like it could introduce lots of F+ indeed. How about adding a check option to disable this?

macournoyer (approved):

PR #232, opened 2021-03-25
Bump min Liquid version to 5.0.1

### LINKED ISSUE #218: echo tag seems to interfere with UnusedAssign
So after converting a block of Liquid to the new `liquid` and `echo` tag, theme-check seems to think my variables are unused.

image

macournoyer:
To include fix for echo tag

Fixes #218


macournoyer:
Build will be fixed by #233

charlespwd (approved):

PR #237, opened 2021-03-25
Fix document links from different root

### LINKED ISSUE #231: Snippet Document Links don't account for root config
I was super psyched for this update. I was actually planning on creating an extension to do exactly this.

Regrettably, there seems to be a small issue where it doesn't account for the `root` config setting, meaning the snippets directory is incorrect.

`root: src/theme`

image
Censored part is the Project Root directory.

charlespwd:
Fixes #231

Had to refactor a little bit here. Made it clearer that the expected API for read/write on Storage is relative_paths. Also made it so the handler was properly using relative paths when calling the completion and document link engines.

Also:
- Added integration tests for document link (with/without root attribute)
- Refactored all the storage.root.join(...).to_s -> storage.path(...) in tests
- Also made the urls in diagnostics use `file://` instead of `file:`. Since I believe `file://` is the standard. (I'm being redirected in a browser anyway.)

### DIFF
--- a/lib/theme_check/file_system_storage.rb
+++ b/lib/theme_check/file_system_storage.rb
@@ -12,7 +12,7 @@ module ThemeCheck
     end
 
     def path(relative_path)
-      @root.join(relative_path)
+      @root.join(relative_path).to_s
     end
 
     def read(relative_path)
--- a/lib/theme_check/in_memory_storage.rb
+++ b/lib/theme_check/in_memory_storage.rb
@@ -6,32 +6,29 @@
 # as a big hash already, leave it like that and save yourself some IO.
 module ThemeCheck
   class InMemoryStorage < Storage
-    def initialize(files = {}, root = nil)
+    def initialize(files = {}, root = "/dev/null")
       @files = files
-      @root = root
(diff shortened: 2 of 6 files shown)


charlespwd:
Dat flaky test 😡 :fire:

charlespwd:
Flaky test fixed in #238

lib/theme_check/file_system_storage.rb
```diff
@@ -12,7 +12,7 @@ def initialize(root, ignored_patterns: [])
     end
 
     def path(relative_path)
-      @root.join(relative_path)
```
macournoyer, line 15: Is this really needed? Would prevent using any `Pathname` methods on this later.

macournoyer (approved):
LGTM!

lib/theme_check/file_system_storage.rb
charlespwd: Actually I did it for consistency with the other one and because I was using + in the unit tests. I definitely can change both to paths!

charlespwd:
Ooops fat finger. I'll add add my PR fixes in another PR.

PR #246, opened 2021-04-07
Complete all the things :tada:

charlespwd:
This PR adds:

- Global object completion from everywhere:
  - Liquid Variables
  - Liquid Filter arguments
  - if/unless/elsif statements
  - for statements
  - case/when statements
  - assign statements
  - cycle statements
  - render statements
  - echo statements + filter arguments
  - assign statements + filter arguments
- Adds _attribute_ completion everywhere (same list)
- Fixes a bugs in the filter completion provider
- Fixes a couple of bugs in the filter completion provider
- Adds complete tags

There's too much in this PR. It's a draft. There's still a lot of refactoring left to do. This code is ugly!! (but it works)


charlespwd:
I'm splitting this PR into smaller ones. I do feel like every bullet point can be committed separately.

Issue #181, opened 2021-02-25
Complete object properties


charlespwd:
1. We would need a list of drops and their attributes
2. We would need to figure out on what drop you are

PR #253, opened 2021-04-08
Object Completions Everywhere

### LINKED ISSUE #182: Complete object properties inside arguments
As a follow up on #181, it would be nice to have completions on the following scenarios:
- In `{% if object.$property %}`
- In `{% assign name = product.$property %}`
- In `{% liquid %}` tags
- In `{% else object.$property %}`

This seems kind of hard but it might be feasible to just chip at it one at a time. Didn't fully think this through.

charlespwd:
Fixes #182

This PR adds the following:

- A VariableLookupFinder class whose job is to find a potential Liquid::VariableLookup under the cursor inside the following places:
  - liquid variables + filter _arguments_:
    - `{{ here | replace: here, and_here | img_url: class: and_also_here }}`
  - liquid if/unless/elsif statements
    - `{% if here %}`
    - `{% elsif here %}`
    - `{% unless here %}`
  - for/tablerow statements
    - `{% for thing in here %}`
    - `{% tablerow thing in here %}`
  - case/when statements
    - `{% case here %}`
    - `{% when here %}`
  - cycle statements
    - `{% cycle here, here, and_here %}`
  - render statements
    - `{% render 'snippet', thing: here, thang: and_here %}`
  - echo statements + filter arguments
    - `{% echo here | replace: here, and_here | img_url: class: and_also_here %}`
  - assign statements + filter arguments
    - `{% assign var = here | replace: here, and_here | img_url: class: and_also_here %}`
  - `{% liquid %}` statements

    ```liquid
    {% liquid
      if here
        echo here | replace: here, and_here | img_url: class: and_also_here
      elsif here
        assign var = here | replace: here, and_here | img_url: class: and_also_here
      endif

      unless here
        render 'snippet', thing: here, thang: and_here
      endunless

      for thing in here
        cycle here, here, and_here
      endfor

      case here
      when here
      endcase
    %}
    ```
- Changes the implementation of the ObjectCompletionProvider to find variable lookups and complete them if we're on a lookup without sublookups

### 🎥 DEMO  👉: https://screenshot.click/08-05-ycrlb-81yq4.mp4


lib/theme_check/language_server/variable_lookup_finder.rb
```diff
@@ -0,0 +1,263 @@
...
+        is_liquid_unless = is_liquid_tag && last_line.match?(/^\s*unless/)
+        is_unless_tag = markup =~ /\A{%\s*unless/im
+        return empty_lookup if (is_unless_tag || is_liquid_unless) && ends_with_spaces
+        markup += '{% endunless %}' if is_unless_tag
+        markup += "\n endunless" if is_liquid_unless
+
+        # elsif statements
+        is_elsif_tag = markup =~ /\A{%\s*elsif/im
+        is_liquid_elsif = is_liquid_tag && last_line.match?(/\A\s*elsif/)
+        return empty_lookup if (is_elsif_tag || is_liquid_elsif) && ends_with_spaces
```
charlespwd, line 127: Might want ends_in_potential_lookup_position here instead.

lib/theme_check/language_server/variable_lookup_finder.rb
```diff
@@ -0,0 +1,263 @@
...
+
+        # for statements
+        is_for_tag = markup =~ /\A{%\s*for/im
+        is_liquid_for = is_liquid_tag && last_line.match?(/^\s*(for)/)
+        return empty_lookup if (is_for_tag || is_liquid_for) && ends_with_potential_lookup_position
+        markup += "{% endfor %}" if is_for_tag
+        markup += "\n endfor" if is_liquid_for
+
+        # closing liquid tag
+        markup += "\n%}" if is_liquid_tag
```
charlespwd, line 160: Wonder if it's worth it to refactor this into something a little more intelligible. I'm not sure moving the code will be any better but it's worth a try.

test/language_server/variable_lookup_finder_test.rb
```diff
@@ -0,0 +1,186 @@
...
+        assert_can_lookup('{% echo  ', '')
+        assert_can_lookup("{% echo \n\t", '')
+        assert_can_lookup('{% echo %}', '', -2)
+        assert_can_lookup('{% echo cart', 'cart')
+        assert_can_lookup('{% echo cart %}', 'cart', -3)
+        assert_can_lookup('{% echo "foo" | replace: ', '')
+        assert_can_lookup('{% echo "foo" | replace: var', 'var')
+        assert_can_lookup('{% echo "foo" | replace: var1, ', '')
+        assert_can_lookup('{% echo "foo" | replace: var1, var2', 'var2')
+        assert_can_lookup('{% echo "foo" | replace: attr: var', 'var')
```
charlespwd, line 62: Don't see why we can't use assert_can_lookup_tag here.

test/language_server/variable_lookup_finder_test.rb
```diff
@@ -0,0 +1,186 @@
...
+
+      def test_can_lookup_cycle_statements
+        assert_can_lookup_tag("cycle cart.error", "cart.error")
+        assert_can_lookup_tag("cycle '', cart.error", "cart.error")
+      end
+
+      def test_can_lookup_for_statements
+        assert_can_lookup_tag("for p in ", "")
+        assert_can_lookup_tag("for p in cart", "cart")
+        assert_can_lookup_tag("for p in cart.error", "cart.error")
```
charlespwd, line 109: Might be worth adding a test for `reversed`

test/language_server/variable_lookup_finder_test.rb
```diff
@@ -0,0 +1,186 @@
...
+
+        assert_can_lookup(<<~LIQUID, 'cart', -1)
+          {% liquid
+            assign foo = blob
+            assign bar = cart
+        LIQUID
+      end
+
+      def test_can_lookup_table_row_statements
+      end
```
charlespwd, line 145: Oh yeah still gotta do that one.

lib/theme_check/language_server/variable_lookup_finder.rb
```diff
@@ -0,0 +1,263 @@
...
+        # - end at cursor position
+        #
+        # That way, we'll have a partial liquid variable that
+        # can be parsed such that the "last" variable_lookup
+        # will be the one we're trying to complete.
+        markup = content[start_index..end_index]
+
+        # Early return for incomplete variables
+        return empty_lookup if markup =~ /\s+$/
+        return if markup =~ /((?<!\[)['"]|\]|\[)$/
```
charlespwd, line 51: Should name that regex. That for square brackets thing. Should also make sure that this also happens in liquid tags.

lib/theme_check/language_server/variable_lookup_finder.rb
```diff
@@ -0,0 +1,316 @@
+# frozen_string_literal: true
+
+module ThemeCheck
+  module LanguageServer
+    module VariableLookupFinder
+      extend self
+
+      UNCLOSED_SQUARE_BRACKET = /\[[^\]]*\Z/
+      UNCLOSED_SINGLE_QUOTE = /'[^']*\Z/
+      UNCLOSED_DOUBLE_QUOTE = /"[^"]*\Z/
```
charlespwd, line 10: Not using these.

data/shopify_liquid/tags.yml
```diff
@@ -8,6 +8,8 @@
 - cycle
 - decrement
 - echo
+- else
```
macournoyer, line 11: I'm wondering, should be put all the `endif`, `endcomment` here so we get completion for those too?

And `where` too (from watching your demo).

lib/theme_check/language_server/variable_lookup_finder.rb
```diff
@@ -0,0 +1,314 @@
...
+        return unless cursor_is_on_liquid_tag_lookup_position(content, cursor)
+
+        markup = parseable_markup(content, cursor)
+        return empty_lookup if markup == :empty_lookup_markup
+
+        template = Liquid::Template.parse(markup)
+        current_tag = template.root.nodelist[0]
+
+        case current_tag.tag_name
+        when "if", "unless"
```
macournoyer, line 131: Would `when` work here too?

lib/theme_check/shopify_liquid/tag.rb
```diff
@@ -8,6 +8,38 @@ module Tag
 
       def labels
         @tags ||= YAML.load(File.read("#{__dir__}/../../../data/shopify_liquid/tags.yml"))
+          .to_set
+      end
+
+      # Add magic regex methods for all the liquid tags
+      # - if_tag => /\A\{\%-?\s*if/
+      # - if_liquid_tag => /^\s*if/m
+      def method_missing(method)
```
macournoyer, line 17: Why not a plain method? `find_tag`?
`method_missing` can lead to confusion.

macournoyer (commented):
Will continue reviewing tomorrow. 👏

lib/theme_check/language_server/variable_lookup_finder.rb
charlespwd: When is gets parsed as `case`. (I grab the when from the contents)

data/shopify_liquid/tags.yml
charlespwd: Yeah I have something coming in the fix for #181. Could put that in here too.

charlespwd: https://github.com/Shopify/theme-check/commit/369c5abb686c3e5221438aeb9e0d3c274584a9cf

I also make it so you can complete them but only when they make sense.

lib/theme_check/shopify_liquid/tag.rb
charlespwd: I wrote it this way to reduce noise in the code that calls it + make sure all the regexes are following the same pattern (supports both `{%` and `{%-`).

It's already pretty verbose:

```
is_if_tag = markup =~ ShopifyLiquid::Tag.if_tag
is_if_liquid_tag = is_liquid_tag && last_line =~ ShopifyLiquid::Tag.if_liquid_tag
```

But it's true that this might not be much different from

```
is_if_tag = markup =~ ShopifyLiquid::Tag.tag_regex('if')
is_if_liquid_tag = is_liquid_tag && last_line =~ ShopifyLiquid::Tag.liquig_tag_regex('if')
```

I'll change it. I guess I wanted to play with those metaprogramming chops I learned from Why's poignant guide :p

lib/theme_check/language_server/variable_lookup_finder.rb
```diff
@@ -0,0 +1,322 @@
...
+        # Now we go to hack city... The cursor might be in the middle
+        # of a string/square bracket lookup. We need to close those
+        # otherwise the variable parse won't work.
+        markup += "'" if markup.count("'").odd?
+        markup += '"' if markup.count('"').odd?
+        markup += "]" if markup =~ UNCLOSED_SQUARE_BRACKET
+
+        variable = variable_from_markup(markup)
+
+        variable_lookup_for_liquid_variable(variable)
```
charlespwd, line 92: Is this logic that different than the one used by the tag? I'm using Variable.new directly but I think that if I didn't strip the start and closed it appropriately, I could have all the logic in one.

macournoyer: It depends where in tags and what tag. For the `if` conditional for eg. it's not the same. It's a `Condition` IIRC.

macournoyer (approved):
Looks very solid. Great stuff! 👏

charlespwd:
Just simplified it even more by removing all the is_liquid_tag logic + last_line hacks. Instead I turn the last_line of the liquid tag into a proper tag and assume everything is a proper tag. 

Works and removes a lot of complexity!

https://github.com/Shopify/theme-check/pull/253/commits/e1604383662e2e756bde5ee09a90fc11a7bcf8b6

PR #268, opened 2021-04-21
theme-check-disable fixes for regex based checks

### LINKED ISSUE #242: theme-check-disable does not work for ImgWidthAndHeight
In _snippets/product-card-list.liquid_ i have

```
<div class="list-view-item__image-wrapper product-card__image-wrapper">
      {% comment %}theme-check-disable{% endcomment %}       <--- works
      {% unless product.featured_media == null%}
      {% comment %}theme-check-enable{% endcomment %}
        {% comment %}theme-check-disable{% endcomment %}     <---- does not work
        <img class="list-view-item__image" src="{{ product.featured_media | img_url: '95x95', scale: 2 }}" alt="{{ product.featured_media.alt }}">
        {% comment %}theme-check-enable{% endcomment %}
      {%endunless%}
</div>
```
When i run `theme-check .` i still get for this file this error:
```
snippets/product-card-list.liquid:23: error: ImgWidthAndHeight: Missing width and height attributes.
	<img class="list-view-item__image" src="{{ product.featured_media | img_url: '95x95', scale: 2 }}" alt="{{ product.fe...
```

charlespwd:
Fixes #242

(and more...)

## Approach used

- Refactor DisabledChecks so that it tracks ranges for which checks were disabled
- Refactor Visitor so that it doesn't disable checks from executing but rather clean up offenses that were disabled. This is so that Regex checks that operate on on_document can be ignored.
- Move the PositionHelper + Position logic from Offense into a Position class that can be reused in Node.
- Add node position helper functions ({start,end}_{index,line,column})
- Have the checks themselves hold their offenses (as opposed to having a global offense array in the analyzer)

## Also in this PR

- Added better handling of offenses when line_numbers, markup or template.source are not available / incompatible.


lib/theme_check/disabled_check.rb
```diff
@@ -0,0 +1,39 @@
+# frozen_string_literal: true
+
+# This class keeps track of checks being turned on and off in ranges.
+# We'll use the node position to figure out if the test is disabled or not.
```
macournoyer, line 4: Brilliant! That's a great idea. Works for everything 👏

lib/theme_check/json_file.rb
```diff
@@ -20,6 +20,10 @@ def relative_path
       @relative_pathname ||= Pathname.new(@relative_path)
     end
 
+    def source
+      @source ||= @storage.read(@relative_path)
+    end
+
```
macournoyer, line 26: Can you update `load!` bellow to use this new memoized `source`.

test/position_helper_test.rb
```diff
@@ -0,0 +1,68 @@
+# frozen_string_literal: true
+require "test_helper"
+
+module ThemeCheck
+  module LanguageServer
+    class PositionHelperTest < Minitest::Test
+      def setup
+        @helper = Object.new
+        @helper.extend(PositionHelper)
```
macournoyer, line 9: You could just `include PositionHelper` in this class, and use the methods without any receiver.

test/position_helper_test.rb
```diff
@@ -0,0 +1,68 @@
...
+      end
+
+      def test_convert_from_row_column_to_index
+        assert_equal(content.index('a'), @helper.from_row_column_to_index(content, 0, 0))
```
macournoyer, line 13: I like that you renamed line to row. So row is a 0-index, and line 1-index.

lib/theme_check/position_helper.rb
```diff
@@ -0,0 +1,37 @@
+# frozen_string_literal: true
+# Note: Everything is 0-indexed here.
+
+module ThemeCheck
+  module PositionHelper
```
macournoyer, line 5: Why split in two between Position and PositionHelper? How do you determine where to add a new position related method?

Maybe just a comment would make it clear.

macournoyer: Also, `bounded` is used in `Position` which breaks the decoupling.

lib/theme_check/position_helper.rb
```diff
@@ -0,0 +1,37 @@
...
+      safe_index = bounded(0, index, content.size - 1)
+      lines = content[0..safe_index].lines
+      row = lines.size - 1
+      col = lines.last.size - 1
+      [row, col]
+    end
+
+    def bounded(a, x, b)
+      [a, [x, b].min].max
+    end
```
macournoyer, line 35: Suggestion: you could use a range here:

```suggestion
    def bounded(a, range)
      [range.begin, [x, range.end].min].max
    end
```

`bounded(x, 1..3)`

test/test_helper.rb
```diff
@@ -89,5 +98,34 @@ def after_node(node)
...
+        matches(source, @regex).each do |match|
+          add_offense(
+            "Regex found #{match[0]}.",
+            node: node,
+            markup: match[0],
+            line_number: source[0...match.begin(0)].count("\n") + 1
+          )
+        end
+      end
+    end
```
macournoyer, line 129: Is there really value in adding these here, they are only used in `DisabledChecksTest`?

And these bleed into `DocumentationTest`.

If you nest them under `DisabledChecksTest`, they will be isolated to that test.

macournoyer (approved):
Amazing! Only suggestions. It works great!

🔥 🔥 🔥 🔥

test/test_helper.rb
charlespwd: Actually, I think you are right. The tests initially were using TracerCheck which _is_ reused in other places. So I put those there as I was changing things. I like that idea. I'll do that.

lib/theme_check/position_helper.rb
charlespwd: Right but I think it would make the calls slightly more difficult to comprehend:

I find the following slightly less cute:
```
bounded(line_number - 1, 0..content.lines.sizes - 1)

# vs

bounded(0, line_number - 1, content.lines.sizes - 1)
```

In my mind it's equivalent to

```
a <= x <= b
```

lib/theme_check/position_helper.rb
charlespwd: PositionHelper is used in the LSP in a different context (to convert row/col info from the client into index. 2D -> 1D)

Moreover:

* PositionHelper is concerned about travelling from 1D to 2D and vice versa
* Position is concerned about stuff with a needle, a haystack and a line_number

Maybe I'm naming the second one poorly? Signature:

```
Position.new(needle, content, line_number)
```

test/test_helper.rb
charlespwd: Actually it's because they depend on LiquidCheck (which does the CheckTracking). I did the offense << Offense.new manually and extend Check directly with no problems.

jmignac (approved):
🔥 Excellent stuff!

PR #269, opened 2021-04-21
Rescue errors thrown by remote_asset_file

charlespwd:
Sometimes the size check doesn't work. It's OK to ignore them.


macournoyer (approved):
Other potential exceptions: https://github.com/Shopify/storefront-renderer/blob/master/lib/exceptions.rb#L25

jmignac (approved):

PR #273, opened 2021-04-23
[WIP] Object Attribute Completions Everywhere

### LINKED ISSUE #181: Complete object properties

charlespwd:
Fixes #181

TODO:

- [ ] Split object.yml and "common_drops.yml"
- [ ] Remove undefined objects from object.yml (product, image, etc.)
- [ ] Annotate checkout objects (where applicable)
- [ ] Add deprecated attributes somewhere so we don't complete them


lib/theme_check/language_server/completion_providers/filter_completion_provider.rb
```diff
@@ -37,6 +37,7 @@ def partial(content, cursor)
         partial_match = matches(content, NAMED_FILTER).find do |match|
           match.begin(1) <= cursor && cursor < match.end(1) + 1 # including next character
         end
+        return '' if partial_match.nil?
```
charlespwd, line 40: That commit should be in its own PR

Annafis187 (approved):

PR #275, opened 2021-04-23
Include operators to space inside braces check

jmignac:
From https://github.com/Shopify/theme-tools/issues/447

- Added: check for consistent spaces around symbol comparison and logical operators.
- Fixed: space after `:` was not accusing offense if the next character was a string quote.
- Fixed: markup on missing space before filter pipeline offense was only pointing to the pipe, and not to the complete match.


macournoyer (approved):
Excellent idea!

PR #277, opened 2021-04-26
Complete end* tags

charlespwd:


data/shopify_liquid/tags.yml
```diff
@@ -2,28 +2,28 @@
 - assign
 - break
 - capture
-- case
```
macournoyer, line 5: Wooo! Nice 👏

lib/theme_check/language_server/completion_providers/tag_completion_provider.rb
```diff
@@ -6,7 +6,9 @@ class TagCompletionProvider < CompletionProvider
       def completions(content, cursor)
         return [] unless can_complete?(content, cursor)
         partial = first_word(content) || ''
-        ShopifyLiquid::Tag.labels
+        labels = ShopifyLiquid::Tag.labels
+        labels += ShopifyLiquid::Tag.end_labels if partial.start_with?('end')
```
macournoyer, line 10: I'm curious why you need the `start_with?` here. Why not return all possible completions, and let the rest of the completion engine deal which one is relevant?

macournoyer (approved):

lib/theme_check/language_server/completion_providers/tag_completion_provider.rb
charlespwd: Yeah. I was on the fence with this one. Wanted to hide the 10ish end\* options if you were scrolling down.

charlespwd: I'll remove that. It's superfluous and debatable.

PR #283, opened 2021-05-07
Add doc for auto-correction of SpaceInsideBraces

macournoyer:
It's the only check that has auto-correction for now. This at least shows what auto-correct does.


shainaraskas (approved):
perfect!

PR #290, opened 2021-05-17
Bump version

macournoyer:
I realized I forgot to update the changelog in the previous release. And maybe last release should have been a minor bump because of Object completion?

Anyways, updated the change log. Anything else missing?


macournoyer:
Updated the release instructions so I don't forget again 🙈

Issue #284, opened 2021-05-11
Different results with different Ruby versions


dzschille:
I use Bitbuckets pipelines to check our theme on every push. Today i saw three different results with Ruby versions 2.7.0 to 2.7.2. This is the build i run:

```
pipelines:
    default:
        - parallel:
            - step:
                name: Run Theme-Kit to check templates and more
                image: ruby:2.7.2
                script:
                    - gem install theme-check
                    - theme-check .
```

This is the result:

```
+ gem install theme-check
Successfully installed liquid-5.0.1
Successfully installed nokogiri-1.11.3-x86_64-linux
Building native extensions. This could take a while...
Successfully installed nokogumbo-2.0.5
Successfully installed theme-check-0.8.1
4 gems installed

+ theme-check .
Checking . ...
sections/timer.liquid:20: style: SpaceInsideBraces: Space missing before '}}'.
	<h1 id="timer-headline" style="color: {{ section.settings.text_color}}">{{ section.settings.title }}</h1>
	                                        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
168 files inspected, 1 offenses detected, 1 offenses auto-correctable
```

Then i switched the image to ruby:2.7.1 and the result is now:

```
+ gem install theme-check
Successfully installed liquid-5.0.1
Successfully installed nokogiri-1.11.3-x86_64-linux
Building native extensions. This could take a while...
Successfully installed nokogumbo-2.0.5
Successfully installed theme-check-0.8.1
4 gems installed

+ theme-check .
Checking . ...
168 files inspected, 0 offenses detected, 0 offenses auto-correctable
```

An when i switched the image to ruby:2.7.0 theme-check doesn't even work:

```
+ gem install theme-check
Successfully installed liquid-5.0.1
Successfully installed nokogiri-1.11.3-x86_64-linux
Building native extensions. This could take a while...
Successfully installed nokogumbo-2.0.5
Successfully installed theme-check-0.8.1
4 gems installed

+ theme-check .
/usr/local/bundle/gems/liquid-5.0.1/lib/liquid/tokenizer.rb:31:in `split': invalid byte sequence in US-ASCII (ArgumentError)
	from /usr/local/bundle/gems/liquid-5.0.1/lib/liquid/tokenizer.rb:31:in `tokenize'
	from /usr/local/bundle/gems/liquid-5.0.1/lib/liquid/tokenizer.rb:11:in `initialize'
[...]

PR #291, opened 2021-05-17
Fix disabling checks that emit offences in `on_end`

macournoyer:
### Problem

Discovered while trying to disable `UndefinedObject`. It didn't work because this checks adds the offenses during `on_end`, at the end of analysis. And we kept the disabled checks per template.

### Solution

Share the same `DisabledChecks` instance across all templates.

Also fix support for disabling offenses without any position (`UnusedSnippet`).

### DIFF
--- a/lib/theme_check/analyzer.rb
+++ b/lib/theme_check/analyzer.rb
@@ -19,7 +19,8 @@ module ThemeCheck
         end
       end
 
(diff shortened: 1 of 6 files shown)


macournoyer:
Also appears to fix the inconsistent results in #284

PR #297, opened 2021-05-21
Add ContentForHeaderModification check to prevent relying on the content of CFH

### LINKED ISSUE #295: Warn when relying on the content of `content_for_header`
For example.:

`{{ content_for_header | split: ... }}` is bad practice, and makes the theme very fragile.

Or also, but less important:

```liquid
{% capture variable %}
  {{ content_for_header }}
{% endcapture %}
```

macournoyer:
Fixes #295

### DIFF
--- a/test/checks/content_for_header_modification_test.rb
+++ b/test/checks/content_for_header_modification_test.rb
@@ -1,0 +1,52 @@
+# frozen_string_literal: true
+require "test_helper"
+
+class ContentForHeaderModificationTest < Minitest::Test
+  def test_reports_use_of_filter
+    offenses = analyze_theme(
+      ThemeCheck::ContentForHeaderModification.new,
+      "layout/theme.liquid" => <<~END,
+        {{ content_for_header | split: ',' }}
+      END
+    )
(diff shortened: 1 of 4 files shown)


test/checks/content_for_header_modification_test.rb
```diff
@@ -0,0 +1,52 @@
...
+  end
+
+  def test_do_not_report_normal_use
+    offenses = analyze_theme(
+      ThemeCheck::ContentForHeaderModification.new,
+      "layout/theme.liquid" => <<~END,
+        {{ content_for_header }}
+      END
+    )
+    assert_offenses("", offenses)
```
charlespwd, line 50: Should we add a test case for `{% liquid %}` blocks and `echo`?

macournoyer: Do you mean check for `echo content_for_header`?

charlespwd: Yes just make sure that still works but not `echo content_for_header | ...`. Probably redundant but I've been bit in the past.

PR #309, opened 2021-06-08
Add ImgLazyLoading check for recommending loading="lazy" attribute

macournoyer:
Add a new check to recommend using the `loading="lazy"` attribute on ``:

```
 <img src="a.jpg">
```

Will report:

```
Add loading="lazy" to defer loading of images at templates/index.liquid:1
```

### DIFF
--- a/lib/theme_check/checks/img_lazy_loading.rb
+++ b/lib/theme_check/checks/img_lazy_loading.rb
@@ -10,13 +10,15 @@ module ThemeCheck
     def on_img(node)
       loading = node.attributes["loading"]&.value&.downcase
       return if ACCEPTED_LOADING_VALUES.include?(loading)
-      class_attribute = node.attributes["class"]&.value
-      return if class_attribute&.split(" ")&.include?("lazyload")
 
-      if loading == "auto"
+      class_list = node.attributes["class"]&.value&.split(" ")
+
+      if class_list&.include?("lazyload")
+        add_offense("Use the native loading=\"lazy\" attribute instead of lazysizes", node: node)
+      elsif loading == "auto"
         add_offense("Prefer loading=\"lazy\" to defer loading of images", node: node)
       else
-        add_offense("Add loading=\"lazy\" to defer loading of images", node: node)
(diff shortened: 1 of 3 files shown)


lib/theme_check/checks/img_lazy_loading.rb
```diff
@@ -0,0 +1,14 @@
+# frozen_string_literal: true
+module ThemeCheck
+  class ImgLazyLoading < HtmlCheck
+    severity :suggestion
+    categories :html, :performance
+    doc docs_url(__FILE__)
+
+    def on_img(node)
+      if node.attributes["loading"]&.value != "lazy"
```
colinbendell: "eager" is also a valid value. Though we probably don't want to encourage this, we should permit it.

test/checks/img_lazy_loading_test.rb
```diff
@@ -0,0 +1,28 @@
+# frozen_string_literal: true
+require "test_helper"
+
+module ThemeCheck
+  class ImgLazyLoadingTest < Minitest::Test
+    def test_no_offense_with_loading_lazy_attribute
+      offenses = analyze_theme(
+        ImgLazyLoading.new,
+        "templates/index.liquid" => <<~END,
+          <img src="a.jpg" loading="lazy">
```
colinbendell, line 10: silly question: is it case sensitive?

lib/theme_check/checks/img_lazy_loading.rb
```diff
@@ -0,0 +1,14 @@
...
+        add_offense("Add loading=\"lazy\" to defer loading of images", node: node)
```
colinbendell, line 10: should also check for 'auto' which was pitched originally but was determined to be a bad idea. we should flag this as well with more nuanced response

colinbendell (changes requested):

test/checks/img_lazy_loading_test.rb
macournoyer: Fixed in d68d170

macournoyer:
The only unsolved issue here is that loading="lazy" shoudn't be used above the fold. Created an issue to figure this out later.

PR #321, opened 2021-06-21
Convert `AssetSizeCSS` to `HtmlCheck`

hannakebedom:
### What does this do?
Convert `AssetSizeCSS` from `LiquidCheck` to `HtmlCheck`. Creates a new `LiquidCheck` for the stylesheet tag  called `AssetSizeCssStylesheetTag`.
### How does it do it?
Makes the `AssetSizeCSS` class a subclass of HtmlCheck. Removes the  processing method and any regex constants.
Removes the need to rely on regexp to parse the HTML. It makes the checks simpler, and more robust.

### DIFF
--- a/config/default.yml
+++ b/config/default.yml
@@ -109,16 +109,22 @@ ParserBlockingJavaScript:
 
 ParserBlockingScriptTag:
   enabled: true
+  ignore: []
 
 AssetSizeJavaScript:
   enabled: false
-  ignore: []
   threshold_in_bytes: 10_000
+  ignore: []
 
 AssetSizeCSS:
   enabled: false
-  ignore: []
   threshold_in_bytes: 100_000
+  ignore: []
+
+AssetSizeCSSStylesheetTag:
+  enabled: false
+  threshold_in_bytes: 100_000
+  ignore: []
 
 ImgWidthAndHeight:
   enabled: true
--- a/docs/checks/asset_size_css_stylesheet_tag.md
+++ b/docs/checks/asset_size_css_stylesheet_tag.md
@@ -1,0 +1,50 @@
+# Check Title (`AssetSizeCSSStylesheetTag`)
+
+The `stylesheet_tag` filter generates a link tag that points to a given stylesheet. This rule exists to prevent large CSS bundles (for speed).
+
+## Check Details
+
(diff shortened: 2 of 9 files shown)


config/default.yml
```diff
@@ -122,6 +122,12 @@ AssetSizeCSS:
   threshold_in_bytes: 100_000
   ignore: []
 
+AssetSizeCssStylesheetTag:
+  enabled: false
+  ignore: []
+  threshold_in_bytes: 100_000
+  ignore: []
```
charlespwd, line 129: Looks like this was a mistake. The ignore: [] twice is redundant. You only need it once (in all the places above and here as well).

docs/checks/asset_size_css_stylesheet_tag.md
```diff
@@ -0,0 +1,50 @@
...
+
+The `threshold_in_bytes` option (default: `100_000`) determines the maximum allowed compressed size in bytes that a single CSS file can take.
+
+## When Not To Use It
+
+This rule is safe to disable.
+
+## Version
+
+This check has been introduced in Theme Check 0.6.0
```
macournoyer, line 41:
```suggestion
This check has been introduced in Theme Check THEME_CHECK_VERSION
```

This will cause the version to be automatically injected on the next version we release.

Although it was mostly extracted from another check, this one is technically new.

lib/theme_check/checks/asset_size_css_stylesheet_tag.rb
```diff
@@ -0,0 +1,38 @@
...
+        return if asset.nil?
+        asset.gzipped_size
+
+      # remote URLs
+      elsif href =~ %r{^(https?:)?//}
+        asset = RemoteAssetFile.from_src(href)
+        asset.gzipped_size
+      end
+    end
+  end
```
macournoyer, line 37: This is duplicated from `AssetSizeCSS`.

You could extract it into the existing `RegexHelpers`, or make a new helper module if you can think of a better name.

macournoyer (commented):
Very nice! Just a few small issues.

I think we can also get rid of these? https://github.com/Shopify/theme-check/blob/303a/lib/theme_check/liquid_check.rb#L9-L19

PR #324, opened 2021-06-23
`DeprecateLazysizes` and `DeprecateBgsizes`

hannakebedom:
### What does this do?
Discourages the use of the `lazySizes` Javascript library(https://github.com/aFarkas/lazysizes) and the `bgSizes `plugin (https://github.com/aFarkas/lazysizes/tree/gh-pages/plugins/bgset).
### How does it do it?
Adds the checks `DeprecateLazysizes` and `DeprecateBgsizes`.
Extracting the `lazySizes` check outside of the current `ImgLazyLoading` check allows us to deprecate lazySizes without it affecting the JS "loading" attribute

### DIFF
--- a/docs/checks/deprecate_lazysizes.md
+++ b/docs/checks/deprecate_lazysizes.md
@@ -1,0 +1,61 @@
+# Deprecate lazySizes (`DeprecateLazysizes`)
+
+[lazysizes](https://github.com/aFarkas/lazysizes) is a common JavaScript library used to lazy load images, iframes and scripts.
+
+## Check Details
+
+This check is aimed at discouraging the use of the lazysizes JavaScript library
+
+:-1: Examples of **incorrect** code for this check:
+
+```liquid
+
+<!-- Reports use of "lazyload" class -->
+<img src="a.jpg" class="lazyload">
+
+<!-- Reports use of "data-srcset" and "data-sizes" attribute. Reports data-sizes="auto" -->
+<img
+  alt="House by the lake"
+  data-sizes="auto"
+  data-srcset="small.jpg 500w,
+  medium.jpg 640w,
+  big.jpg 1024w"
+  data-src="medium.jpg"
+  class="lazyload"
+/>
+
+```
+
+:+1: Examples of **correct** code for this check:
+
+```liquid
+
+<!-- Does not use lazySizes library. Instead uses native "loading" attribute -->
+<img src="a.jpg" loading="lazy">
+
+```
+
+## Check Options
+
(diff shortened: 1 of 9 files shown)


docs/checks/deprecate_lazysizes.md
```diff
@@ -0,0 +1,61 @@
...
+## Check Options
+
+The default configuration for this check is the following:
+
+```yaml
+DeprecateLazysizes:
+  enabled: true
+```
+
+## When Not To Use It
```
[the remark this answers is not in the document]
macournoyer: I would say:

> You should disable this rule if you want to support lazy loading of images in older browser that don't support the [`loading="lazy"` attribute](https://caniuse.com/loading-lazy-attr) yet.

macournoyer (approved):
Wow amazing work all around!

docs/checks/deprecate_lazysizes.md
```diff
@@ -0,0 +1,61 @@
+# Deprecate lazySizes (`DeprecateLazysizes`)
+
+lazySizes is a common Javascript library used to lazy load images, iframes and scripts.
```
charlespwd, line 3:
```suggestion
[lazysizes](https://github.com/aFarkas/lazysizes) is a common JavaScript library used to lazy load images, iframes and scripts.
```

Makes for easier browsing if you don't know what it is. (Also it looks like it's spelled with all lower case)

And JavaScript takes a capital S.

Sorry being super nitpicky :upside_down_smile:

docs/checks/deprecate_lazysizes.md
```diff
@@ -0,0 +1,61 @@
...
+  class="lazyload"
+/>
+
+```
+
+:+1: Examples of **correct** code for this check:
+
+```liquid
+
+<!-- Does not use lazySizes library. Instead uses native "loading" attribute -->
```
charlespwd, line 33:
```suggestion
<!-- Does not use lazysizes library. Instead uses native "loading" attribute -->
```

docs/checks/deprecate_lazysizes.md
```diff
@@ -0,0 +1,61 @@
...
+The default configuration for this check is the following:
+
+```yaml
+DeprecateLazysizes:
+  enabled: true
+```
+
+## When Not To Use It
+
+If you want to use lazySizes, then it's safe to disable this rule.
```
charlespwd, line 49:
```suggestion
If you want to use lazysizes, then it's safe to disable this rule.
```

PR #326, opened 2021-06-25
Fix ParsingHelpers#outside_of_strings handling of empty strings

macournoyer:
Because lookahead was used in the regex, empty strings were missed in parsing, which caused weird issues such as:

```
template.liquid:100: style: SpaceInsideBraces: Space missing after ','.
        value="{{ filter.min_value.value | money_without_currency | replace: '.', '' | replace: ',', '.' }}"
```


charlespwd (approved):
LGTM!

Issue #234, opened 2021-03-25
Add way to extend configs


charlespwd:
The default.yml will get big. Might be good to split that out and allow configs to extend others. For instance, we could have a "nothing.yml" config that makes it so no defaults are merged. Would make it easier to turn all checks off except the ones that you define in your `.theme-check.yml` (without breaking backward compatibility and keeping theme-check evergreen by default.) 

Also throwing it out there:
- performance.yml
- liquid.yml
- theme.yml
- app.yml

macournoyer:
Instead of categories? Or in addition to? Is this because default.yml is used for `--init`?

charlespwd:
I'd say in addition to. It's because the default is _merged_. So an empty .theme-check.yml with nothing in it will actually have all the default checks turned on (or off) depending on the default. 

So, say you want to disable all the checks and only turn enable one, you need to `--init` and `%s/enabled: true/enabled: false/g`. But a new version of theme-check might break your build because the new checks would be enabled by default. Requiring you to add the new checks in your config.yml and explicitly disabling them. 

Whereas if we add support for something like:

```yaml
root: src/theme
extends: nothing
ParserBlockingJavaScript:
  enabled: true
```

You would only run that check forever (without requiring a manual disable)

macournoyer:
What about introducing a special `All` magical check name, similar to Rubocop:

```yaml
All:
  enabled: false

ParserBlockingJavaScript:
  enabled: true
```

We could even introduce other magical names for categories: `Performance`, `Liquid`, etc.

charlespwd:
Haha! Your ruby and my front-end influences are clashing :P 
https://eslint.org/docs/developer-guide/shareable-configs#using-a-shareable-config
https://babeljs.io/docs/en/options#presets
https://www.typescriptlang.org/tsconfig#extends

charlespwd:
That aside, this also works.

macournoyer:
Ah I see! Yeah I don't know enough about the JS ones. Whatever make sense!

charlespwd:
More context on what lighthouse CI does with strict recommendations vs lax defaults:
https://github.com/GoogleChrome/lighthouse-ci/issues/51

PR #333, opened 2021-06-29
Merge changes from the internal repo

catlee:


lib/theme_check/tags.rb
```diff
@@ -169,14 +169,32 @@ class Javascript < Liquid::Raw; end
...
+      def register_tags?
+        @register_tags
+      end
+
+      def register_tag(name, klass)
+        return if Liquid::Template.tags[name]
+        Liquid::Template.register_tag(name, klass)
+      end
+
+      def register_tags!
```
catlee, line 184: there were a lot of rebase conflicts around this area. best to give it a close inspection to make sure I've resolved them correctly.

lib/theme_check/tags.rb
```diff
@@ -169,14 +169,32 @@ class Javascript < Liquid::Raw; end
...
-    Liquid::Template.register_tag('stylesheet', Stylesheet)
+    class << self
+      attr_writer :register_tags
+
+      def register_tags?
+        @register_tags
+      end
+
+      def register_tag(name, klass)
+        return if Liquid::Template.tags[name]
```
macournoyer, line 180: I added this line in the private repo to help Feken, but I see now that it breaks a few things when scanning Dawn. Can you remove it? Or I can push a commit with details if you prefer.

macournoyer (commented):
Looking good, but the line bellow is resulting in lots of false positive for Dawn.

lib/theme_check/tags.rb
catlee: @veken1199 what do we need here for our promote draft checks?

charlespwd:
Fixes #234

lib/theme_check/tags.rb
veken1199: Regardless, we should be fine removing this line of code since Core is depending on this implementation: 
```
return if !register_tags? || (defined?(@registered_tags) && @registered_tags)
```

veken1199: This is what I am currently doing in Core
https://github.com/Shopify/shopify/blob/819186d4f6cecdbdeee2ddef2742d0b2ee9367f1/components/online_store/app/models/online_store/app_extension/theme_check_runner.rb#L7

macournoyer (approved):

veken1199 (approved):

PR #334, opened 2021-06-29
Add checks for theme app extension block js/css

catlee:
This PR adds checks that verify that assets referenced by the `javascript` or `stylesheet` fields of theme app extension blocks are within our configured limits.

They're very similar to the existing AssetSize checks, except they only need to parse the schema and check the appropriate fields if they exist.


docs/checks/asset_size_app_block_javascript.md
```diff
@@ -0,0 +1,64 @@
...
+
+This rule disallows the use of block JavaScript files and external scripts to have a compressed size greater than a configured `threshold_in_bytes`.
+
+:-1: Examples of **incorrect** code for this check:
+```liquid
+<!-- Here assets/chat-widget.js is more than 10KB gzipped. -->
+{% schema %}
+{
+  ...
+  "javascript": "chat-widjet.js"
```
charlespwd, line 19:
```suggestion
  "javascript": "chat-widget.js"
```

docs/checks/asset_size_app_block_javascript.md
```diff
@@ -0,0 +1,64 @@
...
+}
+{% endschema %}
+```
+
+## Check Options
+
+The default configuration is the following.
```
charlespwd, line 26:
```suggestion
The default configuration is the following:
```

docs/checks/asset_size_app_block_javascript.md
```diff
@@ -0,0 +1,64 @@
...
+This includes theme and remote scripts.
+
+## When Not To Use It
+
+When you can't do anything about it, it is preferable to disable this rule using the comment syntax:
+
+```
+{% comment %}theme-check-disable AppBlockAssets{% endcomment %}
+<script src="https://code.jquery.com/jquery-3.6.0.min.js" defer></script>
+{% comment %}theme-check-enable AppBlockAssets{% endcomment %}
```
charlespwd, line 47: I feel like this snippet doesn't apply here? Or that the example is the wrong one.

docs/checks/asset_size_app_block_javascript.md
```diff
@@ -0,0 +1,64 @@
...
+```
+
+This makes disabling the rule an explicit affair and shows that the code is smelly.
+
+## Version
+
+This check has been introduced in Theme Check 1.1.0
```
charlespwd, line 54:
```suggestion
This check has been introduced in THEME_CHECK_VERSION
```

We have a releasing step that changes those automatically for us. This way we don't risk releasing it in 1.0.1 and have the wrong number.

docs/checks/asset_size_app_block_css.md
```diff
@@ -0,0 +1,51 @@
...
+
+This includes theme and remote stylesheets.
+
+## When Not To Use It
+
+This rule is safe to disable.
+
+## Version
+
+This check has been introduced in Theme Check 1.1.0
```
charlespwd, line 42:
```suggestion
This check has been introduced in THEME_CHECK_VERSION
```

docs/checks/asset_size_app_block_css.md
```diff
@@ -0,0 +1,51 @@
...
+
+### `threshold_in_bytes`
+
+The `threshold_in_bytes` option (default: `100_000`) determines the maximum allowed compressed size in bytes that a single CSS file can take.
+
+This includes theme and remote stylesheets.
+
+## When Not To Use It
+
+This rule is safe to disable.
```
charlespwd, line 38: Is it? Will we pick up their config or will we use a hardcoded one for validation?

charlespwd:
Otherwise the code for the checks LGTM!

docs/checks/asset_size_app_block_javascript.md
catlee: perfect, I was wondering what I should do here!

docs/checks/asset_size_app_block_css.md
catlee: Given what I should said, I think we should rephrase this to say something like:

This rule should not be disabled locally since the check will be enforced when promoting versions of the extension.

charlespwd: Gucci :ok_hand:

PR #336, opened 2021-06-30
WIP verify that dawn passes all checks

catlee:


charlespwd:
I feel like what we want is the opposite. This project shouldn't fail a build because the Dawn source changed. But it's true we might want to prevent against theme check blowing up when running on Dawn though. 

FYI, I'm working on shopify theme check CI + GitHub action for dawn. This might be a good opportunity to put something like `--fail-level crash` in my PR to keep this check around.

catlee:
We could check against a specific version of Dawn, as long as we remember to bump it when a new Dawn is released.

macournoyer:
> We could check against a specific version of Dawn, as long as we remember to bump it when a new Dawn is released.

Ah I like that! We could also record the output of theme check on that specific version, and compare it. If it reports offenses, we can make sure it doesn't change.

charlespwd:
> We could check against a specific version of Dawn, as long as we remember to bump it when a new Dawn is released.

It's the opposite I want to prevent, we shouldn't have to fix Dawn in order to add new performance checks. It should be the other way around, upgrade theme check CI and fix the lint errors.

macournoyer:
The thing that is missing, and I was trying to cover w/ my suggestion, is checks that silently stop working. This has happened a couple times already.

Maybe out of scope for this PR, but how do we make sure that a check that reported an offense, still does? Any easy way you can think of?

charlespwd:
> Maybe out of scope for this PR, but how do we make sure that a check that reported an offense, still does? Any easy way you can think of? 

Unit tests? :P Or do you mean it in a way, shouldn't crash? Because that's what I think we should do too. In #713, I added the special exit code 2 (same as ESLint) to represent that theme-check crashed while trying to lint the repo. 

This should test that theme-check returns with an exit code != 2 on Dawn.

macournoyer:
Yeah I guess, this is more of an issue w/ our unit tests, you're right! Nevermind my comment, way out of scope anyways. Lets just check that it doesn't crash (or exit 2)

PR #342, opened 2021-07-05
Autocorrection `DefaultLocale`

hannakebedom:
### What does this do?
Makes sure the theme has a default translation file (for example `locales/en.default.json`).
### How does it do it?
Automatically creates the default translation file if it doesn't already exist.
I chose to add a `create_default_locale_json` method to the corrector class to allow for the creation of the default translation files in a corrector block

### DIFF
--- a/lib/theme_check/checks/default_locale.rb
+++ b/lib/theme_check/checks/default_locale.rb
@@ -7,7 +7,9 @@ module ThemeCheck
 
     def on_end
       return if @theme.default_locale_json
-      add_offense("Default translation file not found (for example locales/en.default.json)")
+      add_offense("Default translation file not found (for example locales/en.default.json)") do |corrector|
+        corrector.create_default_locale_json(@theme)
+      end
     end
   end
 end
--- a/lib/theme_check/theme.rb
+++ b/lib/theme_check/theme.rb
@@ -7,7 +7,8 @@ module ThemeCheck
     LIQUID_REGEX = /\.liquid$/i
     JSON_REGEX = /\.json$/i
 
-    attr_accessor :storage
+    attr_reader :storage
+    attr_writer :default_locale_json
 
     def initialize(storage)
       @storage = storage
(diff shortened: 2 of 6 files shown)


lib/theme_check/theme.rb
```diff
@@ -40,6 +43,11 @@ def default_locale_json
       end
     end
 
+    def create_default_locale_json
+      @default_locale_json = JsonFile.new('locales/en.default.json', @storage)
+      @default_locale_json.touch
```
charlespwd, line 48: Should it be empty or should it at least contain an empty object? Would our platform accept a file that is empty?

```json
{}
```

macournoyer: Yes, it needs to contain `{}`

lib/theme_check/file_system_storage.rb
```diff
@@ -23,6 +23,11 @@ def write(relative_path, content)
       file(relative_path).write(content)
     end
 
+    def touch(relative_path)
+      file(relative_path).dirname.mkpath unless file(relative_path).dirname.directory?
+      file(relative_path).write('')
+    end
```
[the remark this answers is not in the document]
hannakebedom: I decided to modify the `write` method by making it create the file if it doesn't already exist since `create` and `write` would be so similar

lib/theme_check/checks/default_locale.rb
```diff
@@ -7,7 +7,9 @@ class DefaultLocale < JsonCheck
 
     def on_end
       return if @theme.default_locale_json
-      add_offense("Default translation file not found (for example locales/en.default.json)")
+      add_offense("Default translation file not found (for example locales/en.default.json)", line_number: -1) do |corrector|
```
macournoyer, line 10: Does it work if you simply do not pass the `line_number`?

```suggestion
      add_offense("Default translation file not found (for example locales/en.default.json)") do |corrector|
```

hannakebedom: No it doesn't. I decided to modify the `correctable?` method in `offense.rb `so it doesn't rely on `line_number`

DrTeePot (approved):

PR #353, opened 2021-07-07
Changes required to point at main branch

catlee:


macournoyer (approved):
❤️

charlespwd:
Do you know if we can somehow have GitHub redirects for the master->main after a rename? This would still require users to upgrade theme-check in order to have the links working :/

macournoyer:
Apparently, it redirects to the default branch: https://github.com/github/renaming#whats-already-changed

catlee:
image

charlespwd:
Woo! That's perfect. Should try it though. I got burned when trying to do this for theme-check-vscode. The Shopify logo link stopped working.

PR #357, opened 2021-07-08
Add steps to update Theme Check in the CLI when releasing

macournoyer:
This process is starting to be very lengthy... I hope we can automate it at some point.


charlespwd (approved):

macournoyer:
Dup of #358

PR #359, opened 2021-07-08
Adding pagination size check

### LINKED ISSUE #338: Add check for Pagination above `50`
Some stores set their pagination above the default `50` (some reaching `999`). It would be good to have a check which looks for the explicit definition of a pagination above `50`...

```liquid
{%- paginate collection.products by 999 -%}
```

... but also, if feasible (as this is much less likely), to check for the value of `products_per_page` settings data for the off chance the excessive pagination values are defined within the theme settings...

```liquid
{%- paginate collection.products by section.settings.products_per_page -%}
```

taylorsteiger-rad:
Fixes #338

### DIFF
--- a/docs/checks/pagination_size.md
+++ b/docs/checks/pagination_size.md
@@ -17,7 +17,7 @@ This check is aimed at keeping response times low.
 {% paginate collection.products by 12 %}
 ```
 
-Use sizes that are below the `max_size`.
+Use sizes that are integers below the `max_size`, and above the `min_size`.
 
 ## Check Options
 
@@ -27,6 +27,7 @@ The default configuration for this check is the following:
 PaginationSize:
   enabled: true
   ignore: []
+  min_size: 1
   max_size: 50
 ```
 
--- a/docs/checks/pagination_size.md
+++ b/docs/checks/pagination_size.md
@@ -1,0 +1,43 @@
+# Ensure `paginate` tags are used with performant sizes
+
+## Check Details
+
+This check is aimed at keeping response times low.
+
+:-1: Examples of **incorrect** code for this check:
+
+```liquid
+<!-- Using too large of page size -->
+{% paginate collection.products by 999 %}
+```
+
+:+1: Examples of **correct** code for this check:
+
+```liquid
+{% paginate collection.products by 12 %}
+```
+
+Use sizes that are below the `max_size`.
+
+## Check Options
+
+The default configuration for this check is the following:
+
+```yaml
+PaginationSize:
+  enabled: true
+  ignore: []
+  max_size: 50
+```
+
+## When Not To Use It
+
+N/A
+
+## Version
+
+This check has been introduced in Theme Check 1.1.0.
+
+## Resources
+
+[paginate]: https://shopify.dev/api/liquid/objects/paginate
--- a/docs/checks/pagination_size.md
+++ b/docs/checks/pagination_size.md
@@ -37,7 +37,7 @@ N/A
 
 ## Version
 
-This check has been introduced in Theme Check 1.1.0.
+This check has been introduced in Theme Check THEME_CHECK_VERSION.
 
 ## Resources
 
--- a/lib/theme_check/checks/pagination_size.rb
+++ b/lib/theme_check/checks/pagination_size.rb
@@ -1,0 +1,61 @@
+# frozen_string_literal: true
+module ThemeCheck
+  class PaginationSize < LiquidCheck
+    severity :suggestion
+    categories :performance
+    doc docs_url(__FILE__)
+
+    attr_reader :max_size
+
+    def initialize(max_size: 50)
+      @max_size = max_size
+    end
+
+    def on_document(_node)
+      @paginations = {}
+      @schema_settings = {}
+    end
+
+    def on_paginate(node)
+      size = node.value.page_size
+      unless @paginations.key?(size)
+        @paginations[size] = []
+      end
+      @paginations[size].push(node)
+    end
+
+    def on_schema(node)
+      schema = JSON.parse(node.value.nodelist.join)
+
+      if (settings = schema["settings"])
+        @schema_settings = settings
+      end
+    rescue JSON::ParserError
+      # Ignored, handled in ValidSchema.
+    end
+
+    def after_document(_node)
+      @paginations.each_pair do |size, nodes|
+        numerical_size = if size.is_a?(Numeric)
+          size
+        else
+          get_setting_default_value(size.lookups.last)
+        end
+        if numerical_size > @max_size
+          nodes.each { |node| add_offense("Use a smaller pagination size", node: node) }
+        end
+      end
+    end
+
+    private
+
+    def get_setting_default_value(setting_id)
+      setting = @schema_settings.select { |s| s['id'] == setting_id }
+      unless setting.empty?
+        return setting.first['default']
+      end
+      # Setting does not exist, assume it is too big
+      Float::INFINITY
+    end
+  end
+end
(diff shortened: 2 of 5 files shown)


macournoyer (approved):
Wonderful work Taylor! Thanks a lot 🙏

test/checks/pagination_size_test.rb
```diff
@@ -0,0 +1,80 @@
...
+      END
+    end
+
+    def test_flag_use_of_setting_value
+      offenses = analyze_theme(
+        PaginationSize.new(max_size: 50),
+        "templates/index.liquid" => <<~END,
+          <!-- setting size -->
+          {%- paginate collection.products by section.settings.products_per_page -%}
+          {% endpaginate %}
```
nresh, line 59: curious, it seems that you use `{% ... %}` and `{%- ... -%}` with/without the dashes back and forth in these tests - is there a reason for it?  I'm totally new to liquid, but i know it's related to whitespace.  I'm just wondering if this is an important thing to test here?

test/checks/pagination_size_test.rb
```diff
@@ -0,0 +1,80 @@
...
+          {% schema %}
+            {
+                "name": "test",
+                "settings": [
+                    {
+                        "type": "number",
+                        "id": "products_per_page",
+                        "label": "Products per Page",
+                        "default": 51
```
nresh, line 68: it may not be a realistic use case, but what about `0`?  technically it's a Numeric, but we probably should have a message for it

lib/theme_check/checks/pagination_size.rb
```diff
@@ -0,0 +1,61 @@
...
+  class PaginationSize < LiquidCheck
+    severity :suggestion
+    categories :performance
+    doc docs_url(__FILE__)
+
+    attr_reader :max_size
+
+    def initialize(max_size: 50)
+      @max_size = max_size
+    end
```
nresh, line 12: wondering: shouldn't there be a `min_size` as well?  like we would want to warn about anything less than `1` right?

lib/theme_check/checks/pagination_size.rb
```diff
@@ -0,0 +1,61 @@
...
+      if (settings = schema["settings"])
+        @schema_settings = settings
+      end
+    rescue JSON::ParserError
+      # Ignored, handled in ValidSchema.
+    end
+
+    def after_document(_node)
+      @paginations.each_pair do |size, nodes|
+        numerical_size = if size.is_a?(Numeric)
```
nresh, line 39: sorry if I'm missing some assumptions here or if this is handled in some way before this check, but `is_a?(Numeric)` here actually would return true for `-10000`, `-1.5`, `0`, `3.33333` -> wouldn't we want to warn the user about using anything other than a positive integer:
```
if size.is_a?(Integer) &&. size.positive?
```

lib/theme_check/checks/pagination_size.rb
```diff
@@ -0,0 +1,61 @@
...
+        end
+      end
+    end
+
+    private
+
+    def get_setting_default_value(setting_id)
+      setting = @schema_settings.select { |s| s['id'] == setting_id }
+      unless setting.empty?
+        return setting.first['default']
```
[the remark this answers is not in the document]
nresh: also, `.select` returns an array of all matching settings by 'id' - suppose the user sets the settings twice by mistake. wouldn't we want the second setting to overwrite the first?  rather than vice versa?  in that case we would want:
```
return setting.last['default']
```

nresh (commented):
excuse all my comments, nothing serious.  this is literally my first day looking at the project, and shopify/liquid in general, just trying to learn.  I am pretty experienced in ruby however.

lib/theme_check/checks/pagination_size.rb
```diff
@@ -0,0 +1,61 @@
...
+      unless @paginations.key?(size)
+        @paginations[size] = []
+      end
+      @paginations[size].push(node)
+    end
+
+    def on_schema(node)
+      schema = JSON.parse(node.value.nodelist.join)
+
+      if (settings = schema["settings"])
```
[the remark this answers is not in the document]
taylorsteiger-rad: I agree with you, in general, but I don't think that `settings` will changes, which is why there are other places where attributes of the `schema` are hardcoded.

https://github.com/Shopify/theme-check/blob/86646cbd52fdcf81c46a395260f83b706483c468/lib/theme_check/checks/asset_size_app_block_css.rb#L23

lib/theme_check/checks/pagination_size.rb
```diff
@@ -0,0 +1,61 @@
...
+
+    def after_document(_node)
+      @paginations.each_pair do |size, nodes|
+        numerical_size = if size.is_a?(Numeric)
+          size
+        else
+          get_setting_default_value(size.lookups.last)
+        end
+        if numerical_size > @max_size
+          nodes.each { |node| add_offense("Use a smaller pagination size", node: node) }
```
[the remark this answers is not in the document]
taylorsteiger-rad: Good point

test/checks/pagination_size_test.rb
taylorsteiger-rad: These are just variations of the tag in Liquid, so I'm making sure that each variation is caught by the test. This is similar to how other tests check.

test/checks/pagination_size_test.rb
taylorsteiger-rad: I'll include that with the `min_size` update.

lib/theme_check/checks/pagination_size.rb
taylorsteiger-rad: Good point

lib/theme_check/checks/pagination_size.rb
```diff
@@ -0,0 +1,65 @@
...
+        numerical_size = if size.is_a?(Numeric)
+          size
+        else
+          get_setting_default_value(size.lookups.last)
+        end
+        if numerical_size.nil?
+          nodes.each { |node| add_offense("Default pagination size should be defined in the section settings", node: node) }
+        elsif numerical_size > @max_size || numerical_size < @min_size || !numerical_size.is_a?(Integer)
+          nodes.each { |node| add_offense("Pagination size must be a positive integer between #{@min_size} and #{@max_size}", node: node) }
+        end
```
nresh, line 50: very nice.

1 more tiny thing: now that we're explicitly testing for the number being an integer, we should also have a test for the case that that numerical_size is in the right range, but not an integer.  in that case it should also add the offence.

nresh: @taylorsteiger-rad I meant that we needed to add a test in the test suite for this - but it looks like you did this already 👍

taylorsteiger-rad: Ah, I see, that makes sense. Thanks for improving test coverage!

docs/checks/pagination_size.md
```diff
@@ -0,0 +1,44 @@
...
+  max_size: 50
+```
+
+## When Not To Use It
+
+N/A
+
+## Version
+
+This check has been introduced in Theme Check 1.1.0.
```
charlespwd, line 40: This should say `THEME_CHECK_VERSION` instead of 1.1.0, that way our prerelease rake task will change it to the actual version in which this check was introduced.

taylorsteiger-rad: Good to know. Done.

lib/theme_check/checks/pagination_size.rb
taylorsteiger-rad: Resolved with later if check.

lib/theme_check/checks/pagination_size.rb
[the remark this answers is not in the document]
taylorsteiger-rad: @nresh I can make that update. I was assuming that a check in `valid_schema.rb` would prevent duplicate settings, so that we're not having to account for that.

taylorsteiger-rad:
@charlespwd and @macournoyer, is there anything else you need from me before merging?

charlespwd:
Our return from vacation :) Thanks again @taylorsteiger-rad!

PR #364, opened 2021-07-14
feature/windows support

charlespwd:
- Add Windows support to the language server (prep work for [Shopify/theme-check-vscode#5](https://github.com/Shopify/theme-check-vscode/issues/5))
- Add Windows executables

### DIFF
--- a/lib/theme_check/language_server/handler.rb
+++ b/lib/theme_check/language_server/handler.rb
@@ -99,7 +99,7 @@ module ThemeCheck
       end
 
       def path_from_uri(uri)
-        uri&.sub('file://', '')
+        uri&.sub('file://', '')&.sub('/c%3A', '')
       end
 
       def relative_path_from_text_document_uri(params)
--- a/lib/theme_check/language_server/server.rb
+++ b/lib/theme_check/language_server/server.rb
@@ -52,8 +52,19 @@ module ThemeCheck
         response_body = JSON.dump(response)
         log(JSON.pretty_generate(response)) if $DEBUG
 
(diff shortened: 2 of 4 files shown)


jmignac (approved):

lib/theme_check/language_server/handler.rb
```diff
@@ -99,7 +99,7 @@ def text_document_uri(params)
       end
 
       def path_from_uri(uri)
-        uri&.sub('file://', '')
+        uri&.sub('file://', '')&.sub('/c%3A', '')
```
colinbendell, line 102: don't you really want `.chomp`? Or are you expecting \r in the middle of the string?

lib/theme_check/language_server/server.rb
```diff
@@ -52,8 +52,19 @@ def send_response(response)
...
+        # Ruby on Windows turns \n into \r\n. Which means that \r\n
+        # gets turned into \r\r\n. Which means that the protocol
+        # breaks on windows unless we turn STDOUT into binary mode and
+        # set the encoding manually (yuk!) or we do this little hack
+        # here and put \n which gets transformed into \r\n on windows
+        # only...
+        #
+        # Hours wasted: 8.
+        eol = Gem.win_platform? ? "\n" : "\r\n"
+        @out.write("Content-Length: #{response_body.bytesize}#{eol}")
```
colinbendell, line 66: for clarity, I'd suggest breaking out the line termination from the http header name/value pair. also, any concern that the bytesize will be null? If so, I'd want to add a safety here. (I'm guessing not since the response_body is fully buffered?

PR #380, opened 2021-08-04
Autocorrection `UnusedAssign`

hannakebedom:


lib/theme_check/corrector.rb
```diff
@@ -27,5 +27,10 @@ def wrap(node, insert_before, insert_after)
       line.insert(node.range[0], insert_before)
       line.insert(node.range[1] + 1 + insert_before.length, insert_after)
     end
+
+    def remove_line(node)
+      line = @template.full_line(node.line_number)
+      line.clear
```
macournoyer, line 33: Will this end up removing more then we want, say if the line contains HTML?

`<p>...</p>{% assign x = 1 %}`

Worth a test case.

macournoyer (commented):
I think it won't work if the line contains more then the `assign`, but I might be wrong.

lib/theme_check/corrector.rb
hannakebedom: Good point. I'll change my approach to account for this

lib/theme_check/node.rb
```diff
@@ -22,9 +22,7 @@ def markup
     end
 
     def markup=(markup)
-      if tag?
-        @value.raw = markup
-      elsif @value.instance_variable_defined?(:@markup)
```
charlespwd, line 27: Shouldn't do this because of #423. It was bugged but this isn't the way.

PR #381, opened 2021-08-05
Alternating quotes inside HTML attributes

### LINKED ISSUE #362: Quotes inside liquid expressions aren't handled correctly
If I have this code:
```
<img src="{{ "test.png" | asset_url }}">
```

The RemoteAsset check adds an error: "Asset should be served by the Shopify CDN for better performance"

I believe this is because the value of the src attribute is "{{ ", since the double quote inside the liquid expression ends the match. Changing it to `'test.png'` fixes the error.

charlespwd:
Fixes #362

(or is an attempt to fix #362 without breaking performance)


test/html_visitor_test.rb
```diff
@@ -1,52 +1,183 @@
...
+        :on_text, "\n{% endcapture %}\n",
+        :after_document
+      ], @tracer.calls)
+    end
+
+    def test_elements_with_quotes_inside_quotes
+      @attribute_checker = HTMLAttributeIntegrityMockCheck.new
+      @visitor = HtmlVisitor.new(Checks.new([@attribute_checker]))
+      template = parse_liquid(<<~END)
+        <script href="{{ \"a.js\" | asset_url }}" defer></script>
```
catlee, line 58: can we add a test with an example that doesn't escape its quotes? e.g.

`<script href="{{ "c.js" | asset_url }}" defer></script>`

catlee:
I'm not sure what to think about this...

Is there a way we can temporarily remove all the liquid content for the purposes of HTML parsing, and then re-inject it after HTML parsing so that we ultimately keep the original liquid? I can't help but think these substitutions are going to cause us problems down the line.

charlespwd:
> this will fudge things up w/ line & column location for LSP, no?

Maybe? I _did_ leave it to single character swaps for that reason. Thought process being that if the check that runs on some HTML node says the error is between line 5:16 to 6:10, then they would essentially be in the same spot.

It would only _not_ work if you try to index/search for the node value inside the template.source. But you shouldn't do that anyway because you might get the wrong index for a duplicated string.

> Another approach could be to replace all {{...}} & {%...%} with placeholders, then replace the placeholders after parsing. But not sure when to do the last replace: maybe using a visitor on the the Gombo AST, or simply in HtmlNode#value?

I like this idea. Might handle "in the middle of tag" liquid better. Worth a shot.

lib/theme_check/html_node.rb
```diff
@@ -22,35 +22,57 @@ def element?
     end
 
     def children
-      @value.children.map { |child| HtmlNode.new(child, template) }
+      @value.children.map { |child| HtmlNode.new(child, template, @placeholder_values) }
```
macournoyer, line 25: I would memoize all these to avoid replacing each time the method is called.

```suggestion
      @children ||= @value.children.map { |child| HtmlNode.new(child, template, @placeholder_values) }
```

Same w/ `attributes`, `parent`, etc. bellow.

lib/theme_check/html_visitor.rb
```diff
@@ -4,23 +4,40 @@
 
 module ThemeCheck
   class HtmlVisitor
+    include RegexHelpers
     attr_reader :checks
 
+    SINGLE_QUOTE = '\''
+    DOUBLE_QUOTE = '"'
+
```
macournoyer, line 12: Don't need this anymore.

```suggestion
```

macournoyer (approved):
Nice! I think it's pretty solid approach now.

PR #382, opened 2021-08-09
Handle escaped file URIs in language server

### LINKED ISSUE #360: Directories with spaces or special characters doesn't seem to run
### Issue
When trying to run the extension in a workspace where the folder structure contains spaces or special characters (i.e "Documents/Hello There" or "Documents/Hello & There". The theme check extension seems to run, but doesn't produce any errors.

image

When running ```shopify theme check```  it correctly produces errors in the terminal.

When running the extension on another directory with no spaces or special characters, it runs as expected.

image

### Machine Info

Shopify cli version 2.0.1
Extension version 1.0.1

Ubuntu 20.04.2 LTS

**VS Code **
Version: 1.57.1
Commit: 507ce72a4466fbb27b715c3722558bb15afa9f48
Date: 2021-06-17T13:26:56.255Z
Electron: 12.0.7
Chrome: 89.0.4389.128
Node.js: 14.16.0
V8: 8.9.255.25-electron.0
OS: Linux x64 5.8.0-59-generic snap

charlespwd:
Fixes #360

### DIFF
--- a/lib/theme_check/language_server/handler.rb
+++ b/lib/theme_check/language_server/handler.rb
@@ -1,5 +1,7 @@
 # frozen_string_literal: true
 require "benchmark"
+require "uri"
+require "cgi"
 
 module ThemeCheck
   module LanguageServer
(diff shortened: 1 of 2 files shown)


lib/theme_check/language_server/handler.rb
```diff
@@ -98,8 +100,10 @@ def text_document_uri(params)
         path_from_uri(params.dig('textDocument', 'uri'))
       end
 
-      def path_from_uri(uri)
-        uri&.sub('file://', '')&.sub('/c%3A', '')
+      def path_from_uri(uri_string)
+        return if uri_string.nil?
+        uri = URI(uri_string)
+        CGI.unescape(uri.path)
```
charlespwd, line 106: This has the added bonus of handling windows paths for us (`C%3a` becomes `c:` as it should).

Issue #354, opened 2021-07-07
Reject {% javascript %} and {% stylesheet %} liquid tags in app blocks


catlee:
We're going to be rejecting these on the backend, so we should flag these are errors in theme-check as well.

catlee:
We should validate that `target` is valid in the schema as well.

PR #383, opened 2021-08-09
Add check for forbidden tags in theme app extension blocks

karadietz:
Resolves https://github.com/Shopify/theme-check/issues/354

This PR adds a check for the following tags in theme app extension blocks since they will be rejected from the backend:

- `{% javascript %}` 
- `{% stylesheet %}`
- `{% include %}`
- `{% section %}`
- `{% layout %}`

### DIFF
--- a/docs/checks/app_block_valid_tags.md
+++ b/docs/checks/app_block_valid_tags.md
@@ -1,0 +1,41 @@
+# Reject Invalid Tags for Theme App Extension Blocks (`AppBlockValidTags`)
+
+This rule exists to prevent invalid tags from being used in theme app extension blocks which will be rejected when the theme app extension is promoted.
+
+## Check Details
+
+This rule verifies no invalid tags are used for theme app extension app blocks. Invalid tags include:
+
+- `{% javascript %}`
+- `{% stylesheet %}`
+- `{% include 'foo' %}`
+- `{% layout 'foo' %}`
+- `{% section 'foo' %}`
+
+:-1: **Incorrect** code for this check occurs with the use of any of the above tags in theme app extension blocks.
+
+## Check Options
+
+The default configuration for theme app extensions is the following:
+
+```yaml
+AppBlockValidTags:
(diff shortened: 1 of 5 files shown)


docs/checks/app_block_valid_tags.md
```diff
@@ -0,0 +1,41 @@
+# Reject Invalid Tags for Theme App Extension Blocks (`AppBlockValidTags`)
+
+This rule exists to prevent invalid tags from being used in theme app extension blocks which will be rejected when the theme app extension is promoted.
```
veken1199, line 3: >This rule exists to prevent invalid tags
- `{% javascript %}`
 - `{% stylesheet %}`...etc are actually valid tags however they are not allowed to be used in the context of AppBlocks, wondering if the wording can be improved? 


>theme app extension is promoted.

I would keep this platform specific operation `promotion` out of theme_check dictionary.

docs/checks/app_block_valid_tags.md
```diff
@@ -0,0 +1,41 @@
...
+
+```yaml
+AppBlockValidTags:
+  enabled: true
+```
+
+## When Not To Use It
+
+This rule should not be disabled locally since the check will be enforced when
+promoting new versions of the extension.
```
veken1199, line 29: Same commend as above regarding `promotion of theme app extensions`

charlespwd (approved):
Beautiful. Couldn't have done it better. LGTM!

docs/checks/app_block_valid_tags.md
```diff
@@ -0,0 +1,40 @@
+# Reject Invalid Tags for Theme App Extension Blocks (`AppBlockValidTags`)
+
+This rule exists to prevent theme app extension blocks from containing certain tags in their liquid code.
```
charlespwd, line 3: s/certain/forbidden

?

veken1199 (approved):

PR #385, opened 2021-08-11
Autocorrection `MissingRequiredTemplateFiles`

hannakebedom:
### What does this do?
Makes sure all of the required template files in a theme are present.
### How does it do it?
Automatically creates required template files if they don't already exist.
This approach uses methods on the corrector class to create the template files that are missing

### DIFF
--- a/lib/theme_check/checks/missing_required_template_files.rb
+++ b/lib/theme_check/checks/missing_required_template_files.rb
@@ -9,19 +9,33 @@ module ThemeCheck
     doc docs_url(__FILE__)
 
     REQUIRED_LIQUID_FILES = %w(layout/theme)
-    REQUIRED_TEMPLATE_FILES = %w(
-      index product collection cart blog article page list-collections search 404
+
+    REQUIRED_LIQUID_TEMPLATE_FILES = %w(
       gift_card customers/account customers/activate_account customers/addresses
-      customers/login customers/order customers/register customers/reset_password password
-    )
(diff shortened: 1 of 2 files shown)


macournoyer (approved):
Awesome, you're on 🔥 !

I think it's missing for the same reason I mentioned in https://github.com/Shopify/theme-check/pull/388#issuecomment-897837585

lib/theme_check/checks/missing_required_template_files.rb
```diff
@@ -9,19 +9,33 @@ class MissingRequiredTemplateFiles < LiquidCheck
...
       (REQUIRED_LIQUID_FILES - theme.liquid.map(&:name)).each do |file|
-        add_offense("'#{file}.liquid' is missing")
+        add_offense("'#{file}.liquid' is missing") do |corrector|
+          corrector.create(@theme, "#{file}.liquid", "")
+        end
       end
       (REQUIRED_TEMPLATE_FILES - (theme.liquid + theme.json).map(&:name)).each do |file|
-        add_offense("'#{file}.liquid' or '#{file}.json' is missing")
+        add_offense("'#{file}.liquid' or '#{file}.json' is missing") do |corrector|
+          if REQUIRED_LIQUID_TEMPLATE_FILES.include?(file)
```
DrTeePot, line 33: 🙌

DrTeePot (approved):

PR #386, opened 2021-08-11
Correctly highlight substrings inside nodes

### LINKED ISSUE #376: theme-check incorrectly believes there should be spacing in HTML tags
Using the `pagination`'s `next` and `previous` parameters to display HTML shows incorrect suggestions for missing/extraneous spaces. I'm wondering if the line breaks are tripping it out?
```liquid
<div class="pagination__items">
  {{ paginate | default_pagination:
    next: '<i class="icon icon--right-t"></i><span class="icon-fallback__text">Next Page</span>',
    previous: '<i class="icon icon--left-t"></i><span class="icon-fallback__text">Previous Page</span>'
  }}
</div>
```

image

charlespwd:
Fixes #376

If a node's content is the following:

```liquid
{{ var | replace ',',', ' }}`
```

Then, our SpaceInsideBraces check should be able to highlight the second comma inside this string.

Since the markup we're trying to highlight is `,`, and since Liquid only provides _line_ numbers. We needed another mechanism to be able to identify substrings of nodes as offenses.

### DIFF
--- a/lib/theme_check/offense.rb
+++ b/lib/theme_check/offense.rb
@@ -1,11 +1,32 @@
 # frozen_string_literal: true
 module ThemeCheck
   class Offense
+    include PositionHelper
+
     MAX_SOURCE_EXCERPT_SIZE = 120
 
     attr_reader :check, :message, :template, :node, :markup, :line_number, :correction
 
-    def initialize(check:, message: nil, template: nil, node: nil, markup: nil, line_number: nil, correction: nil)
+    def initialize(
+      check:, # instance of a ThemeCheck::Check
+      message: nil, # error message for the offense
+      template: nil, # Template
+      node: nil, # Node or HtmlNode
+      markup: nil, # string
+      line_number: nil, # line number of the error (1-indexed)
+      # node_markup_offset is the index inside node.markup to start
+      # looking for markup :mindblow:.
+      # This is so we can accurately highlight node substrings.
+      # e.g. if we have the following scenario in which we
+      # want to highlight the middle comma
+      #   * node.markup == "replace ',',', '"
+      #   * markup == ","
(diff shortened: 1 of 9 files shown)


lib/theme_check/offense.rb
```diff
@@ -1,11 +1,32 @@
...
+      # looking for markup :mindblow:.
+      # This is so we can accurately highlight node substrings.
+      # e.g. if we have the following scenario in which we
+      # want to highlight the middle comma
+      #   * node.markup == "replace ',',', '"
+      #   * markup == ","
+      # Then we need some way of telling our Position class to start
+      # looking for the second comma. This is done with node_markup_offset.
+      # More context can be found in #376.
+      node_markup_offset: 0,
```
macournoyer, line 27: It's named `node_*`, but technically it could also be linked to the `markup` arg up here. So would it be more accurate to drop the `node_` prefix?

```suggestion
      markup_offset: 0,
```

macournoyer (approved):
Small nit/question about naming.

lib/theme_check/offense.rb
charlespwd: Actually no since it's the opposite. Markup is the needle (the comma), node_markup is the container. So I named it node_markup_offset.

I think it makes more sense in Position.new but at this call site it indeed looks a bit strange.

charlespwd: Right but then it felt weird to me that this is the offset of the markup inside the node.

```
template    = 'hello worldworld'
node.markup = 'worldworld'
markup      = 'o'
node_markup_offset = 6
```

I think it's more accurate as node_markup_offset but it might need to be markup_offset_inside_node_markup

PR #392, opened 2021-08-16
Add --output json option

### LINKED ISSUE #220: Output results as JSON
Hello,

First, great linter! I love it.

Then, I could not found any way to output results as raw format like JSON. 
It could be useful for CI implementation. e.g. to create dedicated issues, PR or other stuff.

What do you think about it?

charlespwd:
Fixes #220


lib/theme_check/json_printer.rb
```diff
@@ -0,0 +1,26 @@
...
+    end
+
+    def offenses_by_path(offenses)
+      offenses
+        .map(&:to_h)
+        .group_by { |offense| offense[:path] }
+        .map do |(path, path_offenses)|
+          {
+            path: path,
+            messages: path_offenses.map { |offense| offense.filter { |k, _v| k != :path } },
```
macournoyer, line 18: Should this keep be named `offenses` instead? It's a bit more than a message.
```suggestion
            offenses: path_offenses.map { |offense| offense.filter { |k, _v| k != :path } },
```

macournoyer (approved):

lib/theme_check/json_printer.rb
charlespwd: Yes. That's an artefact of me ~copying~ being greatly influenced by the JSON format ESLint uses.

PR #407, opened 2021-08-26
Do not flag shopify translations as missing or extra

### LINKED ISSUE #318: MatchingTranslations: Extra translation keys: shopify.
MatchingTranslations check also checks for key: "shopify."

This seems unintended, as the default language will inherit those translations from Shopify, and could be overridden in another language.

charlespwd:
Fixes #318


charlespwd:
>  Why not add all the shopify. keys in a yaml/json file or something?

Oh that works too. I must admit I went pretty quickly on this.

PR #416, opened 2021-08-30
Autocorrection `UnusedSnippet`

hannakebedom:
**What does this do?**
Removes unused snippets

**How does it do it?**
Automatically deletes unused snippets

**Why did you pick this approach over other options?**
This approach uses methods on the corrector class to remove unused snippets


DrTeePot (approved):
Great work!

charlespwd (approved):
:shipit:

PR #422, opened 2021-09-01
WIP Autocorrection `TranslationKeyExists`

hannakebedom:
**What does this do?**
Prevents translation errors in themes.
**How does it do it?**
Adds missing translation keys to locale files
Why did you pick this approach over other options?
This approach uses methods on the corrector class to add key-value pairs to existing JSON locale files


macournoyer (approved):
Nice! Very useful stuff. 👏

test/checks/translation_key_exists_test.rb
```diff
@@ -68,4 +68,49 @@ def test_counts_shopify_provided_translations_as_defined
...
+        {{"samplekey.unknownkey.sample" | translate}}
+        {{"samplekey.example.sample" | translate}}
+      END
+    )
+
+    analyzer = ThemeCheck::Analyzer.new(theme, [ThemeCheck::TranslationKeyExists.new], true)
+    analyzer.analyze_theme
+    analyzer.correct_offenses
+
+    expected = { "key" => "TODO", "nested" => { "key" => "TODO", "unknownkey" => "TODO" }, "samplekey" => { "unknownkey" => { "key" => "TODO", "sample" => "TODO" }, "example" => { "sample" => "TODO" } }, "unknownkey" => "TODO" }
```
charlespwd, line 111: Might be easier to see what's happening by indenting this hash:

```suggestion
    expected = {
      "key" => "TODO",
      "nested" => {
        "key" => "TODO",
        "unknownkey" => "TODO",
      },
      "samplekey" => {
        "unknownkey" => {
          "key" => "TODO",
          "sample" => "TODO",
        },
        "example" => {
          "sample" => "TODO"
        },
      },
      "unknownkey" => "TODO",
    }
```

lib/theme_check/corrector.rb
```diff
@@ -36,5 +36,15 @@ def create_default_locale_json(theme)
       theme.default_locale_json = JsonFile.new("locales/#{theme.default_locale}.default.json", theme.storage)
       theme.default_locale_json.update_contents('{}')
     end
+
+    def add_translation_key(theme, key, value)
```
charlespwd, line 40: Should this be named `add_default_translation_key` considering we're doing it on the default one?

I could see this being expanded in the other direction (you add one to default, but not the others) and we'd probably want `add_translation_key` for that.

DrTeePot (approved):

PR #459, opened 2021-09-16
[Proposal] Concurrent Language Server

charlespwd:
- SpaceInsideBraces fixup for tags without arguments
- Make the language server a concurrent machine


charlespwd:
Interesting factoid:

It took 78 seconds (on main). So the cost of concurrency is a lint that is ~20 second slower.

Curious what it's like on a real machine (not a Windows VM).

I'll try that tomorrow.

In any case, being able to have completions before the 90s mark is a noticeable improvement.

charlespwd:
I think I would rather have had true parallelism here but all libraries I could find used `fork` which you can't use on Windows.

I think we would have to rewrite the handler/server and have some sort of IPC messaging system to have true parallelism. That looks like it would be a lot.

This proposal is maybe 80/20 and isn't all that much harder to understand. Pretty sure you would be able to write handler code without worrying about it too much.

macournoyer:
I love the idea! But I think it would be simpler, and more Ruby-ish, if you'd use `Queue`: https://ruby-doc.org/core-2.5.0/Queue.html. Plus, no extra dependency.

Basically, have a background thread that calls `queue.pop ` (blocking call), and receives the operations. The main thread calls `queue.push` when it receives the LSP messages. Eg.: https://github.com/Shopify/shopify-cli/blob/main/lib/shopify-cli/theme/syncer.rb#L100-L104 (but you only need 1 thread)

That's the approached used in SFR, and the CLI for concurrency.

lib/theme_check/analyzer.rb
```diff
@@ -53,16 +70,23 @@ def analyze_files(files)
         # Call all checks that run on the whole theme
         liquid_visitor = LiquidVisitor.new(@liquid_checks.whole_theme, @disabled_checks)
         html_visitor = HtmlVisitor.new(@html_checks.whole_theme)
-        @theme.liquid.each do |liquid_file|
+        total = total_file_count + files.size
```
macournoyer, line 73: Hum... this is weird. `total_file_count` should already include `files.size`. Are you sure this shouldn't simply be `total_file_count`?

lib/theme_check/language_server/server.rb
```diff
@@ -144,10 +206,16 @@ def read_new_content
       end
 
       def cleanup
+        # Stop listenting to RPC calls
+        Thread.kill(@json_rpc_thread) if @json_rpc_thread&.alive?
```
macournoyer, line 210: I'm not sure how well this will work... If it ends up causing issues, you could call `@in.close` to unblock `@in.gets` similar to unblocking the queues w/ `queue.close`.

But just a gut feeling, might work OK here.

macournoyer (approved):
This is huge! 👏 🔥 

Am I right in saying the  benefits to the end user are:
- Less CPU usage
- More responsive plugin
- Anything else?

lib/theme_check/analyzer.rb
charlespwd: Yeah that part is weird. It's because we're actually running theme check on all files for all the non global checks, and then on the files themselves. So we're running the files twice.

charlespwd:
Not sure about less CPU usage but definitely more responsive plugin!

Also, the progress bar is nice. You're getting a better idea of what's going on / answers the question "is it working?"

lib/theme_check/language_server/server.rb
charlespwd: Reworked and found a bug in the `read_new_content` loop. `@in.read(length + 2)` returns nil if @in is closed. It does not throw an error. This means that if it was impossible to read from STDIN, we entered an infinite loop.

Test code:
```rb
t = Thread.new do
  loop do
    # this code never throws, thus never terminates
    c = $stdin.read(5)
    puts "c", c.inspect
  end
end

sleep 1

$stdin.close

t.join
```

So I added a `raise DoneStreaming if @in.closed?` in there.

Good callout!

PR #472, opened 2021-09-27
Properly shutdown language server

charlespwd:
Turns out the process for closing the server is two messages.

1. A shutdown _request_ sent by the client (expects a response!)
2. An exit _notification_ sent by the client to shut it down

If we're not doing the exit, VS Code doesn't clear the logs. That's why we were getting multiple instances of Theme Check Language Server when restarting the server in VS Code.

Ooops!

https://microsoft.github.io/language-server-protocol/specifications/specification-3-17/#shutdown

Also, I changed `$DEBUG` to `THEME_CHECK_DEBUG`. `$DEBUG` has a special meaning in ruby and it _logs all exception to STDERR_. EVEN WHEN CAUGHT! It took me a while to figure out why the vs code logs were not disappearing. Turns out I had enabled that (because I set THEME_CHECK_DEBUG to true in my environment).


exe/theme-check-language-server
```diff
@@ -4,7 +4,7 @@
 require 'theme_check'
 
 if ENV["THEME_CHECK_DEBUG"] == "true"
-  $DEBUG = true
+  ThemeCheck::THEME_CHECK_DEBUG = true
```
macournoyer, line 7: I think you'll get a Ruby warning here, you're not supposed to re-define constants.

You could either change to `$THEME_CHECK_DEBUG` or add module level method: `ThemeCheck.debug = true`.

lib/theme_check/language_server/bridge.rb
```diff
@@ -68,11 +68,22 @@ def receive_response(id, result)
       # https://microsoft.github.io/language-server-protocol/specifications/specification-current/#responseMessage
       def send_response(id, result = nil, error = nil)
         message = { id: id }
-        message[:result] = result if result
+        message[:result] = result if error.nil?
         message[:error] = error if error
```
macournoyer, line 72:
```suggestion
        if error
          message[:error] = error
        else
          message[:result] = result
        end
```

lib/theme_check/language_server/server.rb
```diff
@@ -135,7 +143,10 @@ def cleanup(status_code)
 
         # Hijack the status_code if an error occurred while cleaning up.
         # 👀 unit tests.
-        return status_code_from_error(@error.pop) unless @error.empty?
+        until @error.empty?
+          code = status_code_from_error(@error.pop)
+          status_code = code if code > status_code
```
macournoyer, line 148: Can you explain the `code > status_code` part in a comment?

macournoyer (approved):
Looks good!

But I would change the `ThemeCheck::THEME_CHECK_DEBUG` to avoid a warning.

exe/theme-check-language-server
charlespwd: Funny story, I moved away from $THEME_CHECK_DEBUG because I got a warning that it wasn't initialized :see_no_evil:

Module level method sounds good!

PR #476, opened 2021-09-28
Make TranslationKeyExists also check section translations

### LINKED ISSUE #473: TranslationKeyExists doesn't check section translations
If i have a section with translations, it will send this error: `"submit" does not have a matching entry in 'locales/en.default.json'.`

```liquid
{{ 'submit' | t }}
```

```json
{
  ...
  "locales": {
    "en": {
      "submit": "Subscribe"
    }
  }
}
```

hannakebedom:
Made the TranslationKeyExists check also look for translation keys in the section schema tag. Resolves #473

### DIFF
--- a/lib/theme_check/checks/translation_key_exists.rb
+++ b/lib/theme_check/checks/translation_key_exists.rb
@@ -5,28 +5,48 @@ module ThemeCheck
     category :translation
     doc docs_url(__FILE__)
 
+    def initialize
+      @schema_locales = {}
+      @nodes = {}
+    end
+
+    def on_document(node)
+      @nodes[node.theme_file.name] = []
+    end
+
     def on_variable(node)
       return unless @theme.default_locale_json&.content&.is_a?(Hash)
-
       return unless node.value.filters.any? { |name, _| name == "t" || name == "translate" }
-      return unless (key_node = node.children.first)
-      return unless key_node.value.is_a?(String)
 
-      unless key_exists?(key_node.value) || ShopifyLiquid::SystemTranslations.include?(key_node.value)
-        add_offense(
-          "'#{key_node.value}' does not have a matching entry in '#{@theme.default_locale_json.relative_path}'",
-          node: node,
-          markup: key_node.value,
-        ) do |corrector|
-          corrector.add_default_translation_key(@theme.default_locale_json, key_node.value.split("."), "TODO")
+      @nodes[node.theme_file.name] << node
+    end
+
+    def on_schema(node)
+      if (schema_locales = JSON.parse(node.value.nodelist.join).dig("locales", @theme.default_locale))
+        @schema_locales = schema_locales
+      end
+    end
+
+    def on_end
+      @nodes.each_pair do |_file_name, file_nodes|
+        file_nodes.each do |node|
+          next unless (key_node = node.children.first)
+          next unless key_node.value.is_a?(String)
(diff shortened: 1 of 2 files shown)


test/checks/translation_key_exists_test.rb
```diff
@@ -39,6 +39,38 @@ def test_ignores_existing_key
...
+        {% schema %}
+          {
+            "locales": {
+              "en": {
+                "submit": "Subscribe"
+              }
+            }
+          }
+        {% endschema %}
+        {{"submit" | t}}
```
charlespwd, line 55: The schema is usually at the bottom of a file. Not the top. I believe this breaks if we flip the order.

lib/theme_check/checks/translation_key_exists.rb
```diff
@@ -5,16 +5,41 @@ class TranslationKeyExists < LiquidCheck
...
+          key_exists?(key_node.value, @theme.default_locale_json.content) || ShopifyLiquid::SystemTranslations.include?(key_node.value),
           "'#{key_node.value}' does not have a matching entry in '#{@theme.default_locale_json.relative_path}'",
+          node,
+          key_node
+        )
+      end
+    end
+
+    def add_as_offense(boolean, message, node, key_node)
+      unless boolean
```
charlespwd, line 40: I don't think this works. We might traverse the node before the schema is defined. (See comment about unit test below).

I believe the right way to do this would be to collect the translation nodes as we're traversing the document.

Something like 
```
def on_variable(node)
  if ...
  if ...
  @nodes << node
end

def on_end
  @nodes.each do |node|
    unless (key_node = node...)
    unless translation_exists?(key_node.value)
    add_as_offense(
      "'#{key_node.value'" not found in #{search_path}"
    )
  end

  def translation_exists?(key)
    key_exists?(key, @theme...) ||
    key_exists?(key, @schema_locales) ||
    SystemTranslations.include?(key)
  end
end
```

Also, I think we could clean that up a bit. If we do `@schema_locales = {}`, then we can call `key_exists?(..., @schema_locales)` even when there wasn't a schema in the page. And I believe this would simplify a lot of the logic.

charlespwd (changes requested):

lib/theme_check/checks/translation_key_exists.rb
```diff
@@ -5,18 +5,31 @@ class TranslationKeyExists < LiquidCheck
     category :translation
     doc docs_url(__FILE__)
 
+    def initialize
+      @schema_locales = {}
+      @nodes = []
```
charlespwd, line 10: Just remembered that Check instances are reused between files. Might want to do something like UnusedAssign and have nodes per file.

```
def initialize
  @nodes = {}
end

def on_document(node)
  @nodes[node.theme_file.name] = []
end

def on_variable(node)
  #...
  @nodes[node.theme_file.name] << node
end

# etc.
```

charlespwd (changes requested):

lib/theme_check/checks/translation_key_exists.rb
charlespwd: Otherwise the @nodes array will keep growing as more files are checked.

hannakebedom: Yea, this makes sense.

DrTeePot (commented):
Almost there!

charlespwd (approved):
LGTM

PR #482, opened 2021-10-13
Edit check docs template

shainaraskas:
## This PR:

- Adjusts the theme check docs template to make it look more similar to the template we use for theme check on shopify.dev.

This is step 1 of a process to try to make sure that the docs on shopify.dev stop falling so far behind the theme check repo. If the templates are similar, we can look at auto-creating issues or prs that use these docs as a starting point


macournoyer (approved):
Awesome! Thanks 🙏

PR #487, opened 2021-10-22
Autocorrection `ParserBlockingJavascript`

hannakebedom:
Automatically adds `defer` to parser blocking script tags.


charlespwd:
This is technically an unsafe change. Since the theme is not guaranteed to still work after this correction was made (e.g. theme that depends on jquery being loaded synchronously so that the inline script tags can assume `$` exists)

I would like us to fix #519 before we push this.

PR #501, opened 2021-11-16
Add image_tag filter support

charlespwd:
- Add support for image_tag filter
- Remove unit test io noise in cli_test.rb

### DIFF
--- a/data/shopify_liquid/filters.yml
+++ b/data/shopify_liquid/filters.yml
@@ -63,11 +63,11 @@ I18nFilter:
 - sentence
 - time_tag
 UrlFilter:
-- image_url
 - stylesheet_tag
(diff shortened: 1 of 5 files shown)


data/shopify_liquid/filters.yml
```diff
@@ -67,6 +67,7 @@ UrlFilter:
 - stylesheet_tag
 - script_tag
 - img_tag
```
DrTeePot, line 69: There's also a new one called `image_url` that has to be used with `image_tag`

DrTeePot: We also deprecated `img_url`

DrTeePot (approved):
Just one extra deprecation 😄

PR #504, opened 2021-11-16
Autocorrection `MissingEnableComment`

hannakebedom:
### What does this do?
When `theme-check-disable` is used in the middle of a theme file, the corresponding `theme-check-enable` comment should gets included.
### What should reviewers focus on?

- Regex used to identify a theme-check enable/disable comment

### DIFF
--- a/lib/theme_check/checks/missing_enable_comment.rb
+++ b/lib/theme_check/checks/missing_enable_comment.rb
@@ -27,7 +27,8 @@ module ThemeCheck
       end
 
       add_offense("#{message} disabled but not re-enabled with theme-check-enable", node: node) do
-        script = node.source.to_enum(:scan, /(?<comment_open>{% comment %}theme-check-disable)(?<checks>.*?)(?<comment_close>{% endcomment %})(?<enclosed>[^\n]*\n[^\n]*\n)/m).map { Regexp.last_match }[-1]
+        node.source.scan(/(?<comment_open>{% comment %}theme-check-disable)(?<checks>.*?)(?<comment_close>{% endcomment %})(?<enclosed>[^\n]*\n[^\n]*\n)/m)
+        # script = node.source.to_enum(:scan, /(?<comment_open>{% comment %}theme-check-disable)(?<checks>.*?)(?<comment_close>{% endcomment %})(?<enclosed>[^\n]*\n[^\n]*\n)/m).map { Regexp.last_match }[-1]
         next if script.nil?
         node.source.insert(script.end(:enclosed), "{% comment %}theme-check-enable#{checks_missing_end_index.include?(:all) ? "" : " #{checks_missing_end_index.join(', ')}"}{% endcomment %}\n")
       end
--- a/lib/theme_check/checks/missing_enable_comment.rb
+++ b/lib/theme_check/checks/missing_enable_comment.rb
@@ -27,10 +27,10 @@ module ThemeCheck
       end
 
       add_offense("#{message} disabled but not re-enabled with theme-check-enable", node: node) do
-        node.source.scan(/(?<comment_open>{% comment %}theme-check-disable)(?<checks>.*?)(?<comment_close>{% endcomment %})(?<enclosed>[^\n]*\n[^\n]*\n)/m)
-        # script = node.source.to_enum(:scan, /(?<comment_open>{% comment %}theme-check-disable)(?<checks>.*?)(?<comment_close>{% endcomment %})(?<enclosed>[^\n]*\n[^\n]*\n)/m).map { Regexp.last_match }[-1]
+        script = node.source.scan(/{% comment %}theme-check-disable.*?{% endcomment %}[^\n]*\n[^\n]*\n/m).last
         next if script.nil?
-        node.source.insert(script.end(:enclosed), "{% comment %}theme-check-enable#{checks_missing_end_index.include?(:all) ? "" : " #{checks_missing_end_index.join(', ')}"}{% endcomment %}\n")
+        index = node.source.rindex(script) + script.length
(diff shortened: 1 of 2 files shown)


lib/theme_check/checks/missing_enable_comment.rb
```diff
@@ -26,7 +26,11 @@ def after_document(node)
         checks_missing_end_index.join(', ') + " " + (checks_missing_end_index.size == 1 ? "was" : "were")
       end
 
-      add_offense("#{message} disabled but not re-enabled with theme-check-enable", node: node)
+      add_offense("#{message} disabled but not re-enabled with theme-check-enable", node: node) do |corrector|
+        script = node.source.to_enum(:scan, %r{(?<comment_open>{% comment %}theme-check-disable)(?<checks>.*?)(?<comment_close>{% endcomment %})(?<enclosed>[^\n]*\n[^\n]*\n)}m).map { Regexp.last_match }[-1]
```
hannakebedom, line 30: Gets MatchData object for the last theme-disable comment in the file

lib/theme_check/checks/missing_enable_comment.rb
```diff
@@ -26,7 +26,11 @@ def after_document(node)
         checks_missing_end_index.join(', ') + " " + (checks_missing_end_index.size == 1 ? "was" : "were")
       end
 
-      add_offense("#{message} disabled but not re-enabled with theme-check-enable", node: node)
+      add_offense("#{message} disabled but not re-enabled with theme-check-enable", node: node) do
+        script = node.source.to_enum(:scan, /(?<comment_open>{% comment %}theme-check-disable)(?<checks>.*?)(?<comment_close>{% endcomment %})(?<enclosed>[^\n]*\n[^\n]*\n)/m).map { Regexp.last_match }[-1]
+        next if script.nil?
+        node.source.insert(script.end(:enclosed), "{% comment %}theme-check-enable#{checks_missing_end_index.include?(:all) ? "" : " #{checks_missing_end_index.join(', ')}"}{% endcomment %}\n")
```
macournoyer, line 32: I think you'll get the last line of the file if you append `$` at the end of your regex instead.

test/checks/missing_enable_comment_test.rb
```diff
@@ -90,4 +90,78 @@ def test_specific_checks_disabled_and_reenabled_with_all_checks_disabled
...
+    }
+    sources = fix_theme(
+      ThemeCheck::MissingEnableComment.new,
+      "templates/index.liquid" => <<~END,
+        <p>Hello, world</p>
+        {% comment %}theme-check-disable OneCheck, TwoCheck{% endcomment %}
+          {% assign x = 1 %}
+        {% comment %}theme-check-disable TracerCheck, AnotherCheck{% endcomment %}
+          {% assign x = 1 %}
+        {% comment %}theme-check-disable{% endcomment %}
```
macournoyer, line 159: Will it work if I remove this line?

By work I mean, either not correct it at all of still append the proper `{% comment %}theme-check-enable{% endcomment %}`. It should leave the template in an acceptable state.

macournoyer (commented):
A couple questions

lib/theme_check/checks/missing_enable_comment.rb
hannakebedom: I believe appending $ at the end of the regex will match the pattern at the end of the string. In this case, I'm looking to get the MatchData for the last occurrence of the pattern in the file. This might not be at the very end of the file so I don't think $ would work.

macournoyer: It's the `to_enum` part that bothers me. It's creating a Fiber underneath for no reason.
So either `/Z` or `.scan(...).last`

DrTeePot (approved):
Looks good, could there be a test that shows where the comment is inserted? e.g Liquid-to-correct:
```liquid
        <p>Hello, world</p>
        {% comment %}theme-check-disable{% endcomment %}
        {% assign x = 1 %}
        {% assign x = 1 %}
```

macournoyer (approved):
Looking gooood! 👏 🔥 

Can you add one more test case to make sure it doesn't break [_Disable checks for the entire document by placing the comment on the first line_](https://github.com/Shopify/theme-check#disable-checks-with-comments).

test/checks/missing_enable_comment_test.rb
```diff
@@ -90,4 +90,102 @@ def test_specific_checks_disabled_and_reenabled_with_all_checks_disabled
...
+        {% comment %}theme-check-disable TracerCheck, AnotherCheck{% endcomment %}
+        {% assign x = 1 %}
+      END
+    )
+    sources.each do |path, source|
+      assert_equal(expected_sources[path], source)
+    end
+  end
+
+  def test_adds_missing_enable_after_last_check_when_multiple_checks_disabled
```
charlespwd, line 138: What about HTML checks like ImgLazy or width + height?

charlespwd (changes requested):

test/checks/missing_enable_comment_test.rb
```diff
@@ -90,4 +90,151 @@ def test_specific_checks_disabled_and_reenabled_with_all_checks_disabled
...
+    end
+  end
+
+  def test_html_check
+    expected_sources = {
+      "templates/index.liquid" => <<~END,
+        <p>Hello, world</p>
+        {% comment %}theme-check-disable ImgWidthAndHeight{% endcomment %}
+        <img src="a.jpg">
+        {% comment %}theme-check-enable ImgWidthAndHeight{% endcomment %}
```
charlespwd, line 219: I'm not sure the implementation works.
- What if the comment was added a couple of lines up?
- What if there are multiple offenses? e.g. two lines with ``?
- What if the comment is on the same line of the offense?
```
{% theme-check-disable ImgWidthAndHeight %}<img src="a">
<img src="a">
```

hannakebedom: The current implementation adds the closing comment one line after the disable comment regardless of how many offenses there are.

PR #505, opened 2021-11-16
Fix Liquid in HTML parsing

### LINKED ISSUE #502: theme-check both incorrectly places an error and ignores the disable comment.
So I have this code on a project, and it's correctly warning about *RemoteAsset*, however, it's firstly placed incorrectly, _and_ it's ignoring my ignore comments.

```liquid
    <script>
      window.theme = window.theme || {};
      theme.money_format = "{{ shop.money_format }}";
    </script>

    {% comment %}theme-check-disable RemoteAsset{% endcomment %}
    <script src="https://widget.reviews.co.uk/product/dist.js" defer></script>
    <script src="https://widget.reviews.co.uk/rating-snippet/dist.js" defer></script>
    <link rel="stylesheet" href="https://widget.reviews.co.uk/rating-snippet/dist.css" />
    {% comment %}theme-check-enable RemoteAsset{% endcomment %}
```

image

charlespwd:
We had issues with our parsing hack. When the liquid had new lines in
it, the #{id}######## hack messed with the line numbers.

Also learned today that the HTML parser doesn't provide an accurate
.to_html value. It is _built_ and not what we provided in the source.
Therefore the two are not equal. The only way to find the source markup
is to look for the tag starting at that line in the source itself. Which
we do in this PR.

Also implemented start/end position methods for the HtmlNode while we're
at it.

Also added tests for HtmlNode#markup,#line_number and position methods.

Fixes #502

### DIFF
--- a/lib/theme_check/html_node.rb
+++ b/lib/theme_check/html_node.rb
@@ -5,12 +5,75 @@ module ThemeCheck
   class HtmlNode < Node
     extend Forwardable
     include RegexHelpers
+    include PositionHelper
     attr_reader :theme_file, :parent
 
-    def initialize(value, theme_file, placeholder_values = [], parent = nil)
+    class << self
+      include RegexHelpers
+
+      def parse(liquid_file)
+        placeholder_values = []
+        parseable_source = +liquid_file.source.clone
+
+        # Replace all non-empty liquid tags with ≬{i}######≬ to prevent the HTML
+        # parser from freaking out. We transparently replace those placeholders in
+        # HtmlNode.
+        #
+        # We're using base36 to prevent index bleeding on 36^3 tags.
+        # `{{x}}` -> `≬#{i}≬` would properly be transformed for 46656 tags in a single file.
+        # Should be enough.
+        #
+        # The base10 alternative would have overflowed at 1000 (`{{x}}` -> `≬1000≬`) which seemed more likely.
+        #
+        # Didn't go with base64 because of the `=` character that would have messed with HTML parsing.
+        #
+        # (Note, we're also maintaining newline characters in there so
+        # that line numbers match the source...)
+        matches(parseable_source, LIQUID_TAG_OR_VARIABLE).each do |m|
+          value = m[0]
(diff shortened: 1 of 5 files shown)


lib/theme_check/html_node.rb
```diff
@@ -27,23 +90,39 @@ def value
     def children
       @children ||= @value
         .children
-        .map { |child| HtmlNode.new(child, theme_file, @placeholder_values, self) }
+        .map { |child| HtmlNode.new(child, theme_file, @placeholder_values, @parseable_source, self) }
     end
 
     def markup
-      @markup ||= replace_placeholders(@value.to_html)
+      @markup ||= replace_placeholders(parseable_markup)
```
charlespwd, line 97: Can't use @value.to_html because of the following:
```ruby
HtmlNode.parse(<<~HTML).to_html == <<~EXPECTED
  <script 
    src="hello" 
    defer
  >
  </script>
HTML
  <script src="hello" defer=""></script>
EXPECTED
```

Basically, to_html is _built_. It's not from the source.

test/html_node_test.rb
```diff
@@ -0,0 +1,106 @@
+# frozen_string_literal: true
+require "test_helper"
+
+module ThemeCheck
+  class HtmlNodeTest < Minitest::Test
+    def test_markup
```
macournoyer, line 6: Maybe missing test cases with Liquid wrapping attributes:

```
<img
{% if ... %}
class="banner"
{% else %}
class="no-banner"
{% endif %}
>
```

macournoyer (approved):
Looks go to me, but I feel we're missing a few edge cases in the tests. Could only think of one...

test/html_node_test.rb
charlespwd: Test passes :) The only case that wouldn't work is the same case that would break a prettier plugin:

```
{% if div %}
<div>
{% else %}
<span>
{% endif %}

{% if div %}
</div>
{% else %}
</span>
{% endif %}
```

But that was already a problem. And not one I think we can solve :/

PR #512, opened 2021-11-25
Add SchemaJsonFormat check

### LINKED ISSUE #16: Lint/prettify JSON
`{% schema %}` & all `.json` files must contain valid JSON.

See https://shopify.dev/docs/themes/sections#using-section-schema-tags.

We currently parse the JSON in Ruby to check for syntax error. But it would be nice to pass it through a real JSON linter/formatter that would give use line number error and better error messages.

charlespwd:
https://user-images.githubusercontent.com/4990691/143473949-f66a8c62-b8b0-4fb9-bb56-2543407e7c41.mp4

Fixes #16

### DIFF
--- a/lib/theme_check/liquid_node.rb
+++ b/lib/theme_check/liquid_node.rb
@@ -78,6 +78,11 @@ module ThemeCheck
       @inner_markup ||= source[block_start_end_index...block_end_start_index]
     end
 
+    def inner_json
+      return nil unless schema?
+      @inner_json ||= JSON.parse(inner_markup)
+    end
+
     def markup=(markup)
       if @value.instance_variable_defined?(:@markup)
         @value.instance_variable_set(:@markup, markup)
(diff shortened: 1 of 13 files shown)


lib/theme_check/liquid_node.rb
```diff
@@ -166,6 +171,10 @@ def block?
       block_tag? || block_body? || document?
     end
 
+    def schema?
+      @value.is_a?(ThemeCheck::Tags::Schema)
```
macournoyer, line 175: I think that won't work when ThemeCheck is running in Core or SFR

macournoyer (approved):
Looks good! Except for the `schema?` method.

lib/theme_check/liquid_node.rb
charlespwd: Why not? It's our namespace? Didn't we copy the code over?

PR #523, opened 2021-12-01
Add scripts/release <version>

charlespwd:
Automates most of the stuff to create a release.

- Creates branch name
- Runs git changelog with correct arguments
- Runs rake prerelease[$version]
- Commits everything
- Opens a PR for you

I've had this as a git exclude for a while now. I shouldn't the only one to benefit. Makes the release process a tad less time consuming.

### DIFF
--- a/RELEASING.md
+++ b/RELEASING.md
@@ -9,6 +9,8 @@
    rake prerelease[$VERSION]
    ```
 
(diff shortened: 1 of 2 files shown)


Precha24 (approved):
Good, thanks

PR #533, opened 2021-12-17
Add support for `#` inline comment tag

charlespwd:
The `#` inline comment tag is coming. This PR adds support for it in theme-check.

Looking forward to using this :+1::
```
{% # theme-check-disable ParserBlockingJavaScript %}
```

Instead of this :-1: :

```
{% comment %}theme-check-disable ParserBlockingJavaScript{% endcomment %}
```

Currently blocked until Shopify/liquid#1498 is merged and we have a new Liquid release.

### DIFF
--- a/README.md
+++ b/README.md
@@ -115,32 +115,31 @@ See [config/default.yml](config/default.yml) for available options & defaults.
 Use Liquid comments to disable and re-enable all checks for a section of your template:
 
 ```liquid
-{% comment %}theme-check-disable{% endcomment %}
+{% # theme-check-disable %}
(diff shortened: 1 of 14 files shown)


ConduciveMammal:
@charlespwd this seems to have been pushed live now. Seems to work fine on my stores

charlespwd:
Hmm. Should probably bug them. It's still "unreleased" even though we use it in prod. 

https://github.com/Shopify/liquid/blob/master/History.md#features

blanklob:
I still have the issue.

Issue #564, opened 2022-03-17
[ bug ] when `root` is specified, Custom Check rule can't be used via require option


kilgore5:
# desired behavior
By nature of specifying the root option (keeping theme build separate from project root), I'd expect to be able to keep my custom rules outside of this root, so I'd need to be able to specify a path accordingly.

# error replication
```
# ./.config/theme-check.yml

root: dist

require:
  - ../src/lib/theme-check/custom_spaces_inside_braces.rb
```

running `shopify theme check --config .config/.theme-check.yml` then raises error...
```
[LoadError] cannot load such file -- dist/../src/lib/theme-check/custom_spaces_inside_braces.rb (LoadError)
```

more traceback...
```
	10: from /Users/johnroyall/.rvm/gems/ruby-2.7.5/gems/shopify-cli-2.12.0/lib/project_types/theme/commands/check.rb:27:in `call'
	 9: from /Users/johnroyall/.rvm/gems/ruby-2.7.5/gems/theme-check-1.9.2/lib/theme_check/cli.rb:127:in `run'
	 8: from /Users/johnroyall/.rvm/gems/ruby-2.7.5/gems/theme-check-1.9.2/lib/theme_check/cli.rb:111:in `run!'
	 7: from /Users/johnroyall/.rvm/gems/ruby-2.7.5/gems/theme-check-1.9.2/lib/theme_check/cli.rb:111:in `new'
	 6: from /Users/johnroyall/.rvm/gems/ruby-2.7.5/gems/theme-check-1.9.2/lib/theme_check/config.rb:94:in `initialize'
	 5: from /Users/johnroyall/.rvm/gems/ruby-2.7.5/gems/theme-check-1.9.2/lib/theme_check/config.rb:211:in `resolve_requires'
	 4: from /Users/johnroyall/.rvm/gems/ruby-2.7.5/gems/theme-check-1.9.2/lib/theme_check/config.rb:211:in `each'
	 3: from /Users/johnroyall/.rvm/gems/ruby-2.7.5/gems/theme-check-1.9.2/lib/theme_check/config.rb:212:in `block in resolve_requires'
	 2: from /Users/johnroyall/.rvm/gems/ruby-2.7.5/gems/shopify-cli-2.12.0/bin/shopify:9:in `require'
	 1: from /Users/johnroyall/.rvm/rubies/ruby-2.7.5/lib/ruby/2.7.0/rubygems/core_ext/kernel_require.rb:83:in `require'
[...]

PR #565, opened 2022-03-17
Fix requiring custom check when root is specified

kilgore5:
Potential fix for #564 

# Basic issue
I'm running `theme check` for a root `dist` and trying to require a custom check that lives outside the `dist` folder

# Notes:
- The spec I wrote doesn't break either way, I believe because when run from command line like `shopify theme check --config .config/.theme-check.yml` then `ThemeCheck::Config.load_config` gets a relative path passed in and I'm not 100% sure how to simulate that with the spec

### DIFF
--- a/lib/theme_check/config.rb
+++ b/lib/theme_check/config.rb
@@ -209,7 +209,8 @@ module ThemeCheck
 
     def resolve_requires
       self["require"]&.each do |path|
-        require(File.join(@root, path))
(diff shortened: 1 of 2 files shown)


charlespwd (approved):
Not sure I understand the note but the pathname.join().realpath seems like the good approach indeed :sweat_smile:

kilgore5:
@charlespwd much belated thanks for the quick merge!

PR #574, opened 2022-04-13
Fixed broken URL

chrish-d:
This might be a wider problem, but I spotted it here and fixed it.


charlespwd:
Thanks! I'll check if we have this somewhere else.

PR #608, opened 2022-07-18
Fix UnusedAssign false positives from `{% render ... with var %}`

### LINKED ISSUE #582: UnusedAssign reported when using "with" for render
Hi,

Assuming the following Liquid code:

```
{%- case status -%}
  {%- when 'success' -%}
    {%- assign icon = 'success' -%}
  {%- when 'error' -%}
    {%- assign icon = 'error' -%}
  {%- when 'warning' -%}
    {%- assign icon = 'warning' -%}
{%- endcase -%}

<div class="banner banner--{{ status }} {% if text_alignment != 'left' %}justify-center{% endif %}">
  {%- render 'icon' with icon, width: icon_width, height: icon_width -%}
  {{- content -}}
</div>
```

The linter report the "icon" assigned to not be used (which is not the case, it is used in the "with" attribute).

charlespwd:
Fixes #582

Echo of https://github.com/Shopify/liquid/pull/1596 in our SFR port.

### DIFF
--- a/lib/theme_check/tags.rb
+++ b/lib/theme_check/tags.rb
@@ -145,7 +145,7 @@ module ThemeCheck
 
       disable_tags "include"
 
-      attr_reader :template_name_expr, :attributes
+      attr_reader :template_name_expr, :variable_name_expr, :attributes
 
       def initialize(tag_name, markup, options)
         super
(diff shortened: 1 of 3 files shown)


karreiro (approved):
Thank you, @charlespwd!

PR #633, opened 2022-08-27
Fix: IndexError: string not matched

hrdwdmrbl:
Issue: A settings file, for example `{foo: "bar"}` could exist. It could be missing a property "foo.baz". The checker will attempt to write `{foo: {baz: "TODO"}}`. However, the write will fail because the value of "foo" is not a Hash but a String.
The simplest solution is to re-write the value of "foo" to be a Hash `{foo: {}}` and then write the rest of the nested value (in this case, `{baz: "TODO"}`).

Caution: This could be lossy. However, I believe that the assumption is that source control is in place already.


Precha24 (approved):
Good, thanks

Issue #642, opened 2022-09-26
[Intelligent code completion] Introduce the `type_inference` component with base scenarios and an integration test suite


karreiro:
Introduce the initial implementation of the  `type_inference` component covering the base cases handled by `type_predictions` in the `feature/181-object-property-completions` branch.

This task aims to polish the scenarios already covered, and also introduce a test suite covering all scenarios we're targeting.

PR #654, opened 2022-11-07
Introduce the `ObjectAttributeCompletionProvider` module

karreiro:
## Overview

The goal of this PR is fixing https://github.com/Shopify/theme-check/issues/642* and introducing new intelligent code completion capabilities on Theme Check.

I recommend reviewing this PR by following the sections below. I've split it into meaningful commits to approach each module individually and clarify its purpose.

#### 1. Minor adjustments

- c8c7f3c7e5e13d7df67d611a96fc992b3f17b6b - First of all, we slightly adjust a detail in the Markdown documentation
- 911c53a9ae792a0cb748c97e6b63b160d60e2c8 - Then, we introduce the prototype implementation to support built-in types
  - This module still have some TODO elements: https://github.com/Shopify/theme-check/issues/657

#### 2. Setup our new code completion tools

- 52029b5bf134edb612c5028494fb66123d2f829 - Now, we create the `variable_lookup_finder/constants` module to hold some regexes (they will be used in the future commits)
- 753ce9a30a05064377b7167dadf6dd10048c817 - Then, we introduce the main module of the type inference, the `LiquidFixer`
  - It receives a Liquid string and tries to turn it in a valid/parsable Liquid string
  - This module is designed to receive unfinished snippets, as while developers are typing, templates will be probably broken
- 6b458821b1e24f6a954e54a1d806a69e44e1209 - After that, we introduce the `AssignmentsFinder` module
  - This module relies on a Liquid visitor to get all `assign` occurrences
  - We still have one missing detail to finish in this module here: https://github.com/Shopify/theme-check/issues/655
- e27f9dfaa43f85042acffd21e57531a1c09948b - With those elements in place, we can now change the `VariableLookupFinder` to use the `AssignmentsFinder` and the `LiquidFixer` modules
  - With them, the `VariableLookupFinder` can follow variables and get their types based on assignments

#### 3. Improve our code completion suggestions

- 13b37515a63f2f5c922f39ded6c7fafb848dd8b - Cool! We have our new mechanism in place; it's time to use it to provide valuable suggestions
  - But, before that, we need to apply some minor changes
  - The `CompletionEngine` works as an aggregator of all completion providers (each completion provider extends `CompletionProvider`)
  - With that in mind, we're changing the `CompletionProvider` main method from `completions(content, cursor)` to `completions(relative_path, line, col)`
    - The goal of this change is empowering each provider to suggest completion items based on the **file**, the **line**, and the **column** (instead of `content`, which was ~~the current token~~ a subset of the file)
  - Good! We're done with that
    - Now, let's update the existing completion providers and introduce documentation on them as well (as we have it for free now) :)

#### 4. The new `ObjectAttributeCompletionProvider` provider

- da2c38fe51ca3ecfb474a95a74752aa06c115e8 - Ok, in the latest commits, we updated the existing completion providers, it's time to introduce the new 'ObjectAttributeCompletionProvider' provider
  - Our new provider relies in the new `VariableLookupFinder` implementation to provide intelligent code completion for attributes
  - Now, we're pretty much done

#### 5. TODO

- fc749d3db443166a055e71e605ff7f3585b0d5b8 & a324fb326570b60601b78537b73f6bf2eaa34ca9 - These commits just introduce some references about the pending work we have

## Steps to 🎩

- Setup your editor (VS Code, vim, or nvim) with your local Theme Check by following [this guide](https://github.com/Shopify/theme-check/blob/be6d59c69da1ffbda7d2a380a2db6b2bb19e66c1/CONTRIBUTING.md#run-language-server)
- Create a new theme with `shopify theme init`
- Open your new theme to try code completion
- Notice that filters/tags/objects code completion items have documentation now
- Notice that objects support multi-level attribute code completion now:

---

\* I've been calling the features introduced by this PR as "type inference" in https://github.com/Shopify/theme-check/issues/642 issue. But, I've decided to break that concept into sub-components into this PR to clarify what each phase means.


mgmanzella (commented):
absolute 🔥!! only question i have is the same one CP raised [here](https://github.com/Shopify/theme-check/pull/654#discussion_r1018284512)

lib/theme_check/shopify_liquid/source_index/base_entry.rb
```diff
@@ -7,19 +7,19 @@ class BaseEntry
...
         end
 
         def summary
-          hash['summary']
+          hash['summary'] || ''
         end
 
         def description
-          hash['description']
+          hash['description'] || ''
```
Poitrin, line 22: Not sure in which PR you want to tackle this (did not find a linked TODO), so I'll leave it here :-)
The descriptions are still kinda "rough"/unformatted: they contain symbols like …
- `--`
- `> `
- `<br><br>`

14-50-vg84t-eh4ao

Poitrin: Also maybe related to this:
When clicking on links…
14-42-xpxs4-m538o
… VS Code opens an invalid URL:
14-40-dbu2m-lngn4

lib/theme_check/language_server/completion_providers/object_completion_provider.rb
```diff
@@ -3,29 +3,32 @@
 module ThemeCheck
   module LanguageServer
     class ObjectCompletionProvider < CompletionProvider
-      def completions(content, cursor)
-        return [] unless (variable_lookup = variable_lookup_at_cursor(content, cursor))
+      def completions(relative_path, line, col)
```
Poitrin, line 6: I'm not sure how exact the completion provider needs to be (after all, VS Code already displays the error _Expected end_of_string but found dotdot_), so is it okay if completions are also suggested when syntax is erroneous (e.g. `dotdot`)?

14-57-qxm05-lu0l0

Poitrin, line 6: (Also not sure in which PR we will tackle the following issue:)

This is working:
14-07-1fkoq-ggk88

But completion inside an HTML property does not:
14-08-8xiju-ssvdh

lib/theme_check/language_server/completion_engine.rb
```diff
@@ -3,35 +3,23 @@
...
-        token = find_token(buffer, cursor)
-        return [] if token.nil?
-
-        @providers.flat_map do |p|
-          p.completions(
-            token.content,
-            cursor - token.start
-          )
+        @providers.flat_map do |provider|
+          provider.completions(relative_path, line, col)
```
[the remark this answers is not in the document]
karreiro: Thank you, @charlespwd! Excellent point. I've introduced the `CompletionContext` object to avoid IO calls/token-finding.

Thus, now completion providers implement `completions(context)` instead of `completions(relative_path, line, col)` to support that :)

lib/theme_check/language_server/variable_lookup_finder.rb
```diff
@@ -1,51 +1,48 @@
...
       end
 
       private
 
+      def potential_lookup(variable, token)
+        return variable if token.buffer.nil? || token.buffer.empty?
+
+        buffer = token.buffer[0...token.absolute_cursor]
+        lookups = variable.lookups
+        assignments = find_assignments(buffer)
```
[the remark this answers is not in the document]
karreiro: Thanks for raising that, @charlespwd!

I had something similar to the `AssignmentsManager` running on `textDocument/didChange` in mind, still when I started implementing it I noticed some details that weren't looking so good:

- I. `AssignmentsManager` would visit Liquid templates on every file change (regardless that's used for code completion or not)
- II. Also, it would require some index management to handle multiple files
- III. Also, it would be almost required to have some kind of memoization mechanism to avoid re-visiting tokens (as they would happen a bit frequently) \*

With those points in mind, I've moved the `AssignmentsFinder` to the `textDocument/completion` phase, and then we execute it only when it's required.

Do you have any thoughts about that?

<br>

\--

<br>

> Some of the concerns I'd have with the other approach is handling things like for x in list and if else statements. e.g.
> ```liquid
> {% liquid
>  if type == "product"
>     assign x = product
>  elsif some_condition
>     # kind of convoluted example, but this part of the tree 
>     # technically doesn't have a type for `x`, and I'd be worried
>     # about something that looks at all the code until now to break.
>     assign y = x.value
>  endif
> %}
> ```

I believe we may introduced some level of scope support on the `AssignmentsFinder` to handle iterations, so it would depend on the new ✨ `CompletionContext` instead of `buffer`.

With that enhancement, it would get the type of `p` when we can infer the collection type, and also the `last_product` type:

```liquid
{% for p in collections.first.products %}
  {% assign last_product = p %}
  {{ p.title }}
{% endfor %}

{{ last_product.█ }}
```

Still, I don't think we can should cover this kind of scenario (https://github.com/Shopify/theme-check/issues/655), as we'd need SFR/runtime-context to have confidence about the type of `target` (which depends on `some_condition`):

```liquid
{%- liquid
  if some_condition
    assign target = cart
  else
    assign target = product
  endif
%}
{{ target.█ }}
```

Do you have any extra thoughts about that?

Thanks again for raising those points, @charlespwd! :)

lib/theme_check/language_server/completion_engine.rb
```diff
@@ -3,35 +3,27 @@
...
-      def find_token(buffer, cursor)
-        Tokens.new(buffer).find do |token|
-          # Here we include the next character and exclude the first
-          # one becase when we want to autocomplete inside a token
-          # and at most 1 outside it since the cursor could be placed
-          # at the end of the token.
-          token.start < cursor && cursor <= token.end
-        end
+      def context(relative_path, line, col)
+        CompletionContext.new(@storage, relative_path, line, col)
```
mgmanzella, line 26: suuuuper clever 🔥

mgmanzella (approved):
🚀 🚀 🚀

lib/theme_check/language_server/completion_providers/object_completion_provider.rb
karreiro: Actually, I believe that the the completion items, in the screenshot above, are coming from the default VS Code language server.

The completion items provided by our language server have different icons for different kinds of completion items:

image

Also, I believe it's working inside of HTML attributes. Still, please @Poitrin, feel free to re-open this comment if I'm missing something :)

Poitrin: Yeah, I was not able to reproduce the issue I mentioned :)

lib/theme_check/language_server/variable_lookup_finder/tolerant_parser.rb
```diff
@@ -0,0 +1,89 @@
...
+        end
+
+        class Tags
+          module TolerantBlockBody
+            def parse_body(body, tokens)
+              super
+            rescue StandardError
+              false
+            end
+          end
```
charlespwd, line 49: I feel like this deserves an explanation :D

lib/theme_check/language_server/completion_providers/object_attribute_completion_provider.rb
```diff
@@ -0,0 +1,108 @@
...
+
+        [object, generic_type]
+      end
+
+      def partial_property_name(property, variable_lookup)
+        last_property = variable_lookup.lookups.last
+        last_property if last_property != property&.name
+      end
+
+      def property_doc(prop)
```
charlespwd, line 73: Shouldn't this be `to_completion_item`? It's not obvious otherwise that this is where we map to the LanguageServerProtocol spec.

lib/theme_check/language_server/completion_engine.rb
```diff
@@ -11,7 +11,7 @@ def initialize(storage, bridge = nil)
 
       def completions(relative_path, line, col)
         @providers.flat_map do |provider|
-          provider.completions(relative_path, line, col)
+          provider.completions(context(relative_path, line, col))
```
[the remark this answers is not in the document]
karreiro: 🤦 Indeed! Good catch! Fixed :)

lib/theme_check/language_server/variable_lookup_finder/assignments_finder.rb
```diff
@@ -6,48 +6,161 @@ module VariableLookupFinder
...
+          # {%- liquid
+          #   assign var1 = some_value
+          #
+          #   if condition
+          #     assign var1 = another_value
+          #            ^^^^ from here we no longer can guarantee
+          #                 the type of `var1` in the global scope
+          # -%}
+          # ```
+          @global_scope.delete(node.value.to) unless scope.equal?(@global_scope)
```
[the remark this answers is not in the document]
karreiro: Excellent catch! 🚀  Fixed!

lib/theme_check/language_server/variable_lookup_finder/assignments_finder.rb
```diff
@@ -6,48 +6,161 @@ module VariableLookupFinder
...
 
-          visitor(template).visit if template
+          visit(node, @global_scope)
+        end
+
+        def visit(node, scope, level = 0)
+          return if node.type_name == :variable_lookup
+
+          method = :"on_#{node.type_name}"
+          scope = send(method, node, scope) if respond_to?(method)
```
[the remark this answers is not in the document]
karreiro: That would be much clear indeed! 🔥  I've applied this refactoring in a new commit :)

lib/theme_check/language_server/completion_providers/object_attribute_completion_provider.rb
karreiro: Fixed! 🚀

PR #667, opened 2022-11-22
Introduce the `AssignmentsCompletionProvider` to suggest variables

### LINKED ISSUE #659: [Intelligent code completion] Introduce the `AssignmentsCompletionProvider` to suggest variables
Analog to the `ObjectCompletionProvider`, the new `AssignmentsCompletionProvider` will be pretty much the same thing, however it will rely on the `AssignmentsFinder` to suggest variables for users.

In the example below, target must be suggested:
```liquid
{%- liquid
  assign target = cart
%}
{{ tar█ }}
```

The editor already suggests that kind of term by default. However, if we suggest it, we may show the `cart` documentation for the target variable.

Poitrin:
### WHY are these changes introduced?

Fixes #659 

### WHAT is this pull request doing?

Introduce the `AssignmentsCompletionProvider` to suggest variables

### How to test your changes?

1. Make sure that you are on the correct `theme-check` branch.
2. Test with Liquid code similar to this:
```
{%- liquid
  assign product = all_products.first
  assign prod_uct = product
%}
{{ prod█ }}
```
3. When you start typing `prod`, do you see the following suggestions?
23-32-pyhc4-r394p
4. Does `product` only appear _once_? Do `product` and `prod_uct` display the correct description (namely the one about the globally available variable `product`)?

Add the following code below:
```
{%- liquid
  assign image1 = all_products.first.images.first
%}
{{ ima█ }}
```
1. When you type `ima`, does it suggest `image1`?
2. When you type `ima` in the first example (instead of `prod`), does it **not** suggest `image1`?


karreiro (approved):
Thank you, @Poitrin!

PR #671, opened 2022-11-28
Refresh theme-liquid-docs on theme-check startup

### LINKED ISSUE #651: [Intelligent code completion] Introduce module to auto refresh `theme-liquid-docs` on `theme-check`
The process of refreshing the `theme-liquid-docs` must follow something like these phases:

- Check if the current schema is outdated and if this check has been already run today
  - Part of this effort is creating a module to identify if the files are outdated
  - Initial proposal: rely on the latest commit SHA (if we follow this approach, we need to include a SHA at the [release phase](https://github.com/Shopify/theme-check/issues/650))
 
- Download the schema in a temporary directory (asynchronously at the language server startup)
  - Part of this effort is identifying the best approach for downloading it
  - Initial proposal: we shouldn't _depend_ on Git, we may rely Git to download `theme-liquid-docs` and just "fail" gracefully/silently with users that don't have Git (those users will always rely on the bundled version introduced by https://github.com/Shopify/theme-check/issues/650)
 
- Update the `theme-liquid-docs` files in the [/data/shopify_liquid/documentation/](https://github.com/Shopify/theme-check/commit/ee50185d5a2e6285fc49cb9d0d8ede9f3cbddc6d#diff-bc37d034bad564583790a46f19d807abfe519c5671395fd494d8cce506c42947R19) directory
- Reload the `SourceIndex`

mgmanzella:
### Summary

Closes #651  

Refreshes docs async on theme-check startup, after discussing with @karreiro we decided it's out of scope to periodically refresh while theme-check is running

Since this is a medium size PR, I broke this down into 4 commits:
- [Add async download in `SourceManager#refresh` and mechanism to notify `SourceIndex` liquid schema is out of date](https://github.com/Shopify/theme-check/pull/671/commits/b086c47d5f610965c55ff8e686db7f6a6e7899e7)
   - `SourceManager#download_or_refresh` is called in `on_initialized` to download docs async while theme-check loads the existing outdated docs
   - `SourceManager#refresh` downloads updated docs to a temp directory in case the download fails
   - `SourceIndex::FilterState`, `SourceIndex::ObjectState`, and `SourceIndex::TagState` are used to notify `SourceIndex` if a schema is out of date after refresh
   - `SourceManager` marks each liquid schema variable's respective state to "out of date" when updated docs are successfully downloaded
   - `SourceIndex` marks each liquid schema variable's respective state "up to date" when it finishes loading it
- [Tests for SourceIndex state classes and SourceManager#refresh, create helper for shared stubs](https://github.com/Shopify/theme-check/pull/671/commits/fd348e06dc441f73ab137211ebd643d5cbc1e54d)
    - New tests for new state classes and a small refactor to abstract source stubbing into its own helper
- [Cleanup -- fix naming from `test/shopify_liquid/scope_index` to `test/shopify_liquid/source_index/`](https://github.com/Shopify/theme-check/pull/671/commits/63f536336f9672024ac76a4406887776b7ca3da5)
    - Some cleanup since I was already modifying files in this directory, just changes the test directory for source_index class tests 
- [Remove class method mock (Net::HTTP.any_instance), affects other tests](https://github.com/Shopify/theme-check/pull/671/commits/249a66b2f142244ba4fcf4b5bb39cfa46d0c1a2e)
   - Change to fix flaky tests that were failing because of mocking any_instance of Net::HTTP

### Tophatting Steps 
- In your terminal run `rake download_theme_liquid_docs` if you dont have the version of theme-liquid-docs with a revision file
- In your local copy of theme-check, modify SourceManager#refresh to add `sleep(30)` in the thread
- Modify the revision file at `data/shopify_liquid/documentation/latest.json` to change the revision key to "out_of_date_sha"
- Modify the local copy of your docs to change the documentation for a particular object (eg modify `collections` documentation)
- Restart theme-check and open a theme
- Confirm autocomplete still works
- Confirm the docs are downloaded, the revision is updated in latest.json, and the docs are updated in theme-check when using autocomplete


lib/theme_check/shopify_liquid/source_index.rb
```diff
@@ -11,26 +11,49 @@
 require_relative 'source_index/return_type_entry'
 require_relative 'source_index/tag_entry'
 
+require_relative 'source_index/base_state'
+require_relative 'source_index/filter_state'
+require_relative 'source_index/object_state'
+require_relative 'source_index/tag_state'
+
```
karreiro, line 18: We believe we might centralize the require calls here: [`lib/theme_check/shopify_liquid.rb`](https://github.com/Shopify/theme-check/blob/language-server-docs-main/lib/theme_check/shopify_liquid.rb) - I've noticed this pattern after the merge of the `SourceIndex` PR 😅

lib/theme_check/shopify_liquid/source_manager.rb
```diff
@@ -9,31 +15,68 @@ module SourceManager
...
         end
       end
 
-      def local_path(file_name)
-        documentation_directory + "#{file_name}.json"
+      def refresh
+        return unless refresh_needed?
+
+        @refresh_thread = Thread.new do
+          tmp_dir = Dir.mktmpdir
```
karreiro, line 36: Personally, I'd pass a block (`Dir.mktmpdir { |tmp_dir| ... }`), so `tmpdir` takes care of removing the temporary directory for us :)

lib/theme_check/shopify_liquid/source_manager.rb
```diff
@@ -9,31 +15,68 @@ module SourceManager
...
         required_file_names.each do |file_name|
-          download_file(local_path(file_name), remote_path(file_name))
+          download_file(local_path(file_name, destination), remote_path(file_name))
         end
       end
 
-      def local_path(file_name)
-        documentation_directory + "#{file_name}.json"
+      def refresh
+        return unless refresh_needed?
```
karreiro, line 33: I've noticed some delay in the first code completion. We might avoid that by moving this check into the thread below:

```ruby
def refresh
  @refresh_thread = Thread.new do
    if refresh_needed?
      tmp_dir = Dir.mktmpdir

      begin
        download(Pathname.new(tmp_dir))
        FileUtils.cp_r("#{tmp_dir}/.", documentation_directory)

        mark_all_indexes_outdated
      ensure
        FileUtils.remove_entry(tmp_dir)
      end
    end
  end
end
```

It's a bit uglier :( but I believe it worths as it unblocks the language server :)

karreiro (commented):
Thank you, @mgmanzella! Excellent stuff 🚀 

I've left just some minor comments, please, let me know what you think about them :)

lib/theme_check/shopify_liquid/source_manager.rb
mgmanzella: i suspect ill have to do something [like this](https://github.com/Shopify/theme-check/blob/main/lib/theme_check/remote_asset_file.rb#L34-L37) where i use `Net::HTTP.get(uri)` directly

lib/theme_check/shopify_liquid/source_manager.rb
```diff
@@ -1,81 +1,110 @@
...
-
-      def download_or_refresh_files
-        if has_required_files?
-          # TODO: https://github.com/Shopify/theme-check/issues/651
-          #
-          # Refresh files if they exist locally
-        else
-          download
+    class SourceManager
+      REQUIRED_FILE_NAMES = %i(filters objects tags latest).freeze
```
Poitrin, line 10: /nit – I have "Format on save" activated and it transforms me this line into
`REQUIRED_FILE_NAMES = [:filters, :objects, :tags, :latest].freeze`
Not sure where this rule is defined though, Rubocop doesn't complain :)

lib/theme_check/shopify_liquid/source_index.rb
```diff
@@ -3,50 +3,52 @@
...
         private
 
         def load_file(file_name)
-          read_json(SourceManager.local_path(file_name))
+          read_json(local_path(file_name))
         end
 
-        def read_json(path)
-          SourceManager.download_or_refresh_files
+        def local_path(file_name)
```
Poitrin, line 41: `SourceManager.local_path` only builds the path to `file_name`.
`local_path` in _this_ class however also might download some files.
Therefore, I would rename this method.

lib/theme_check/shopify_liquid/source_manager.rb
```diff
@@ -1,81 +1,110 @@
...
 
-        URI.parse(remote).open.read
-      end
+          revision(local_latest_content) != revision(remote_latest_content)
+        end
 
-      def destination_exist?
-        documentation_directory.exist?
-      end
+        def revision(json_content)
```
Poitrin, line 66: I'm not sure about `!` conventions, but if this method can raise an error, we might rename it to `revision!` ? :D

Poitrin (approved):
🎩 ing passed ✅ 
I just have some nitpicks/remarks :)

lib/theme_check/shopify_liquid/source_manager.rb
mgmanzella: so im hesitant to do this because raising an error here isn't necessarily dangerous in theme-check? refresh erroring out will prevent doc updates but the user will still be able to use theme-check (and even code completion) 😄 

also for internal errors, we need to raise in order to send a message to the client that something went wrong (they're caught and sent [here](https://github.com/Shopify/theme-check/blob/main/lib/theme_check/language_server/server.rb#L142-L149) but the handler keeps chugging along 😎 )

lib/theme_check/shopify_liquid/source_index.rb
Poitrin: I think that `local_path!` is a better name :+1:

I was just wondering if we could write
```
def load_file(file_name)
  SourceManager.download unless SourceManager.has_required_files?
  read_json(SourceManager.local_path(file_name))
end
```
instead and remove the `local_path!` method?

Or do we need it for testing purposes?

Poitrin (approved):
Thanks for your changes and feedback, LGTM!

Issue #679, opened 2022-12-12
[Intelligent Code Completion] Suggestions drop down stuck in `Loading...` state when using assign


mgmanzella:
**Describe the bug**

When trying to assign a variable in liquid, suggestions for an object never load. This only seems to happen when using intelligent code completion with `assign` and then it causes autocomplete to fail for other objects afterwards 

**Expected**

When using assign, suggestions drop down is loaded with relevant functions associated with the object 

**Actual**

The suggestions drop down never loads with assign

To reproduce:
- Open a theme and a liquid file
- Type `{{ product.collections }}`
- In a new line type `{% assign p = █ %}` and at the cursor press `Ctrl - Space`
- Confirm no suggestions are listed and the menu stays in a `Loading...` state
- Confirm you can't trigger anymore intelligent code complete suggestions in the file

![Screenshot 2022-12-12 at 2 25 33 PM](https://user-images.githubusercontent.com/18431672/207136116-8ece5b8f-9ec6-419f-90aa-b9c6a7891c7e.png)

**Stack trace**
None that I can see in the debug log file

**Debugging information**
- OS [e.g. Windows, Mac, Linux] - Mac OS Ventura Version 13.0
- Version (output of `theme-check --version` in a terminal) - 1.12.0

PR #680, opened 2022-12-13
Fix scenarios involving circular references

karreiro:
Fixes https://github.com/Shopify/theme-check/issues/679

The `VariableLookupFinder` was falling in an infinite loop in scenarios involving circular references, which was affecting future completion requests.

**Before**:

**After**:


Poitrin (approved):
Fantastic fix!
I can confirm that your changes have solved the issue.
→ LGTM!

Issue #676, opened 2022-12-09
[UX] Suggest _deprecated_ filters for input type `variable` (so prominently)?


Poitrin:
**Is your feature request related to a problem? Please describe.**
Developers might wonder why they get suggested _deprecated_ filters for input type `variable` for integers and other object types.

<img width="400" alt="Screenshot 2022-12-08 at 12 25 09 PM" src="https://user-images.githubusercontent.com/18431672/206521436-6c73ef6c-a80e-4048-b911-d9c473cb3126.png">

**Describe the solution you'd like**
_In general, it should be a solution that most of our developers prefer._

_Suggestion:_ We update the docs so that the 4 aforementioned deprecated filters get the input type `image` instead. That way, they are only suggested when working with `image`s like `article.image`.

**Describe alternatives you've considered**
1. Hide all deprecated filters. (Or at least deprecated `variable` filters.)
2. Sort deprecated variable suggestions to the bottom

**Additional context**
- https://github.com/Shopify/theme-check/pull/669#issuecomment-1343995170

PR #688, opened 2022-12-15
Suggest deprecated completion items at the end of the list

karreiro:
Fixes https://github.com/Shopify/theme-check/issues/676

This PR handles deprecated completion items into the proper completion providers.

demo_

### DIFF
--- a/lib/theme_check/language_server/completion_provider.rb
+++ b/lib/theme_check/language_server/completion_provider.rb
@@ -38,6 +38,16 @@ module ThemeCheck
           },
         }
       end
+
+      def deprecated_hash(entry)
+        return {} unless entry
+        return { sortText: entry.name } unless entry.deprecated?
+
+        {
+          tags: [CompletionItemTag::DEPRECATED],
+          sortText: "~#{entry.name}",
+        }
+      end
     end
   end
 end
--- a/lib/theme_check/language_server/completion_providers/object_completion_provider.rb
+++ b/lib/theme_check/language_server/completion_providers/object_completion_provider.rb
@@ -22,12 +22,14 @@ module ThemeCheck
 
       private
 
-      def object_to_completion(object)
-        content = ShopifyLiquid::Documentation.object_doc(object)
+      def object_to_completion(object_name)
+        object = ShopifyLiquid::SourceIndex.objects.find { |entry| entry.name == object_name }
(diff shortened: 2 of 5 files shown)


lib/theme_check/language_server/completion_providers/object_completion_provider.rb
```diff
@@ -22,12 +22,14 @@ def partial(variable_lookup)
 
       private
 
-      def object_to_completion(object)
-        content = ShopifyLiquid::Documentation.object_doc(object)
+      def object_to_completion(object_name)
```
Poitrin, line 25: In `def completions`, you call `ShopifyLiquid::Object.labels` which runs `SourceIndex.objects.map(&:name)`.
And then here, you call `ShopifyLiquid::SourceIndex.objects.find`.

Why not call `SourceIndex.objects` directly then, `.select { |w| w.name.start_with?(partial(variable_lookup)) }`, and give the whole object as a parameter to `object_to_completion`, so that no `.find` operation needs to be made?

lib/theme_check/language_server/completion_provider.rb
```diff
@@ -38,6 +38,16 @@ def doc_hash(content)
           },
         }
       end
+
+      def deprecated_hash(entry)
```
Poitrin, line 42: When I read `deprecated_hash`, it sounds as if the hash is _always_ deprecated / as if the method would _always_ mark the hash as deprecated.
But this method feels more like `format_hash`, `format_hash_to_language_server_protocol`, `transform_hash`, …
WDYT?

Poitrin (commented):
I'm pretty sure that this will result in a much nicer DX!

I just have some comments :)

lib/theme_check/language_server/completion_provider.rb
karreiro: Excellent catch!! 🔥🔥🚀

Now, we're no longer missing `discount`:

image

lib/theme_check/language_server/completion_providers/object_completion_provider.rb
karreiro: Excellent catch!! 🔥🔥🚀

Now, we're no longer missing `discount`:

image

karreiro:
Thank you, @Poitrin! Excellent refactoring suggestions, I've applied them in a new commit :)

PR #692, opened 2022-12-16
Fix theme liquid docs 17

Poitrin:


karreiro (approved):
Thank you, @Poitrin! 🚀 Nice stuff :)

PR #703, opened 2023-01-13
Avoid strict dependency between `Shopify/theme-check` unit tests and `Shopify/theme-liquid-docs`

karreiro:
This PR removes `assert_equal(exp, act)` in favor of `assert_operator(act, :>=, min_exp)`.

Thus, `Shopify/theme-check` unit tests will no longer break when new types get documented at `Shopify/theme-liquid-docs`.


mgmanzella (approved):
thank you @karreiro! 🚀

PR #713, opened 2023-02-17
The `--update-docs` messages should use the STDERR

karreiro:
The `--update-docs` messages should use the STDERR [1], as it should inform users but not interfere with scripting when piping (re: https://github.com/Shopify/theme-check-action/pull/19#discussion_r1109019038).

[1] https://clig.dev/#the-basics


Poitrin (approved):
LGTM! (and makes sense!)

PR #723, opened 2023-03-07
Improve rule for lazy loading to prevent developers from overusing it

krzksz:
Shopify has an over lazy loading problem, where theme developers add `loading="lazy"` to images that should actually be loaded as fast as possible. You can read more about this problem here: https://calendar.perfplanet.com/2022/lazy-loading-lcp-images-why-does-this-anti-pattern-happen/

This PR aims to improve the message behind the image lazy loading check to:

- Provide cleaner guideline what can be done to make the warning disappear.
- Give more information on how to decide if `eager` or `lazy` value should be used in a certain scenario.

This is a part of a wider initiative better described by @siakaramalegos here: https://github.com/Shopify/online-store/issues/784

### DIFF
--- a/docs/checks/img_lazy_loading.md
+++ b/docs/checks/img_lazy_loading.md
@@ -8,17 +8,19 @@ Very often, webpages contain many images that contribute to data-usage and how f
 
 _Quoted from [MDN - Lazy loading][mdn]_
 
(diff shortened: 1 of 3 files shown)


siakaramalegos:
LGTM

karreiro (approved):
Thank you for this PR and for linking those valuable resources, @krzksz!

PR #725, opened 2023-04-03
[TAE] add section context to app blocks

patrickracicot:
### What are you trying to accomplish?

Although initially the `section` object was not a valid local-global variable inside app-block context, we are re-introducing it which will allow app-block developers to leverage it to consume section rendering APIs.

Important to note that only the ID property will be exposed.


mmorissette (approved):

shainaraskas:
does this check / the check's documentation also need to be updated?

https://github.com/Shopify/theme-check/blob/main/docs/checks/app_block_valid_tags.md

PR #746, opened 2023-09-29
Clarify Documentation

dbolesta:
Cleans up a bit of the `img_width_and_height` doc to make things clearer.

- Removes a reference definition that was not being used anywhere in the document
- Adds a reference link to clarify the "padding-top hack", which is not common knowledge, and provides further context to what a "padding-top hack" is and how it works.
- Fixes an error where the word "container" was places inside the small css snippet in error.`overflow: hidden container` makes it seems like `container` is a CSS value, when really what is means is "a container with `overflow: hidden` CSS rule applied to it".

### DIFF
--- a/docs/checks/img_width_and_height.md
+++ b/docs/checks/img_width_and_height.md
@@ -52,7 +52,7 @@ ImgWidthAndHeight:
 
 There are some cases where you can avoid content-layout shift without needing the width and height attributes:
 
-- When the aspect-ratio of the displayed image should be independent of the uploaded image. In those cases, the solution is still the padding-top hack with an `overflow: hidden container`.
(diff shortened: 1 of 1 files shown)


lukeh-shopify:
👋🏻 Hi @dbolesta! Thanks for your contribution. Unfortunately the Ruby version of theme check is no longer receiving dev support and is in the process of being deprecated. 

Our recommendation is to use the latest version of theme check via the latest version of the Shopify CLI, or via the Shopify Liquid Visual Studio Code extension. 

If there are docs that need clarifying over there, please raise an issue or make a change over in the [CLI](https://github.com/Shopify/cli) or [theme tools](https://github.com/Shopify/theme-tools) repositories. Thanks!

